PR #2, opened 2015-04-29
Nested comments

awead:


fatmilktv:
@awead , not to hijack this conversation, and I know you're really busy,

but please review the following:

present_terms  ( @presenter, @presenter.ter \ ms - [:title] )                                    do |r, term|
where                 ( provider: auth.provider, uid: \  auth.uid )           . first_or_create   do |user|

I wish to understand what the - \ is doing, in this context. Files attached on request.

One is a Hydra standard form, and one appears to be a complex list comprehension which might, at least this once, be better off unpacked. Maybe even really, really slowly? Thoughts?

Evidently the - is not in core, and rubylint likewise flags it as a custom method of some sort.

Al.

fatmilktv:
Data_Wrap_Struct() returns a created DATA object. The klass argument is the class for the DATA object. The mark argument is the function to mark Ruby objects pointed by this data. The free argument is the function to free the pointer allocation. If this is -1, the pointer will be just freed. The functions mark and free will be called from garbage collector.

@scossu isn't this patently false? There is no Garbage Collector in C.

fatmilktv:
Sorry, I'm going off the clock for this one. Back in a sec.

[generator.txt](https://github.com/aic-collections/aicdams-lakeshore/files/28686/generator.txt)

fatmilktv:
I nominate @cbeer to carry this torch. --Al.

fatmilktv:
Ok, back to work.

awead:
@fatmilktv I'm not sure I understand. I don't see any backslashes. Are you talking about this line: 
https://github.com/aic-collections/aicdams-lakeshore/blob/master/app/views/generic_files/_show_descriptions.html.erb#L3

PR #75, opened 2015-08-07
Changes to masthead, navigation, footer, layout

surreal8:


awead:
The first commit from me shouldn't be there. Commit 2caa22855948b98b9514894998d86572f6c17bf7 is already in master.

I think you need to rebase one more time. From your repo:

```
git checkout master
git pull
git status [to verify you are in sync with your origin]
git checkout aicmods
git rebase master
```

Run `git status` and it should tell you that your branch has diverged, but each should have the same number of commits. Do a force push: `git push origin +aicmods` and that should resolve the issue.

surreal8:
I tried rebase, but no change in that commit showing up and now there's an additional merge commit:
Merge remote-tracking branch 'upstream/master'
Any ideas?

PR #138, opened 2015-12-04
edits to search bar drop down items

surreal8:
@awead here are edits to the search drop-down. For the 'All Assets' option, do templates need to be created like for 'Work' here: https://github.com/aic-collections/aicdams-lakeshore/search?utf8=%E2%9C%93&q=work_path? I just used 'works_path' for the search-option for now as a placeholder.


awead:
@surreal8 this should work now. To search for assets only, we can use the same facet search technique, but I'll need to make a change to GenericFilesController.

PR #149, opened 2015-12-10
changes and additions to work detail page

surreal8:


awead:
We can't merge this because it's breaking tests. If you look at the details link above, it looks like it has to do with the feature tests that show CITI resources. If the changes are correct, then we'll need to update the tests to reflect this.

surreal8:
i'm going to try adding this method to the other presenter files to see if that helps.

PR #163, opened 2015-12-18
Improve relationship list in non-asset view

surreal8:


surreal8:
Looks like I have to edit other files these changes are affecting, will work on this.

surreal8:
@awead do you mind if I make changes to files in the spec folder? Looks like my edits need to be reflected in those files

awead:
@surreal8 nope, not at all. If the spec tests need to change to reflect your changes, then that's what they're there for.

Issue #157, opened 2015-12-17
Finalize available fields in asset edit view


scossu:
- Hide fields that are not in scope for beta 1
- Associate lists with appropriate fields
- Rearrange field order 

The final list of required fields and their order is pending discussion with DEA.

scossu:
List of changes: 
- Hide or remove:
  - Batch UID
  - Contributor
  - Resource created
  - Described By
  - Pref label
  - Resource label
  - Legacy UID
  - Rights
  - Same as
  - Resource updated
  - Icon
  - Document ids
  - Representation ids
  - Pref. repr. ids
  - Transcript
  - Comments
- Rename and rearrange the remaining fields in the following order: 
  - Document Type
  - Title
  - Creator
  - Abstract or Summary
  - Keyword
  - Language
  - Rights Holder
  - Asset Capture Device
  - Compositing -> Image Compositing (list)
  - Light Type -> Image Capture Light Type (list)
  - View -> Image View (list)

surreal8:
Removed fields, moved title to top, need to move Document Type and rename fields

scossu:
That looks good for beta1. You can close the issue as soon as the PR is merged.

scossu:
@awead Already tested in dev. You can deploy to production.

PR #180, opened 2015-12-22

surreal8:


surreal8:
need to edit a couple more files

PR #184, opened 2015-12-23

surreal8:


surreal8:
@awead could you check my edits in edit_generic_file_spec.rb? just want to make sure my deletions are all good.

spec/features/edit_generic_file_spec.rb
```diff
@@ -18,26 +18,5 @@
...
-      fill_in("generic_file[comments_attributes][2][content]", with: comment2)
-      click_button("upload_submit")
-      expect(find_field("generic_file[comments_attributes][0][content]").value).to eql comment0
-      expect(find_field("generic_file[comments_attributes][1][content]").value).to eql comment1
-      expect(find_field("generic_file[comments_attributes][2][content]").value).to eql comment2
-      within("div.comments-editor") { first(:button, "Remove").click }
-      click_button("upload_submit")
-      expect(find_field("generic_file[comments_attributes][0][content]").value).to eql comment1
-      expect(find_field("generic_file[comments_attributes][1][content]").value).to eql comment2
-    end
```
awead, line 41: @surreal8 why are we completely removing this test?

surreal8:
@awead I removed many fields from the the asset edit page, this included the comments field. this test was giving me errors because it was not finding that field i'm assuming?

awead:
With this commit, you've removed the ability to add comments to assets. I'm assuming this is only temporary. @scossu ? if so, then we should mark the test pending to indicate this is functionality that will be available again later.

@surreal8 in the original test, insert this line after the `it` statement and leave the rest of the test intact:

``` ruby
it "supports adding and removing" do
  skip "disabled for beta1 release"
```

Otherwise, if the intention is to in fact remove the ability to add comments to asses, then you can remove this entire file since it only consists of this one test.

surreal8:
@awead @scossu ok, i'll add that line now - thanks!

scossu:
@awead @surreal8 That seems a good temporary fix. Test can be re-enabled where appropriate after the beta release.

PR #186, opened 2015-12-23

surreal8:


surreal8:
@scossu @awead please merge and then i'll close the issue

PR #226, opened 2016-07-19
Adding multilevel doc type menus

surreal8:
This is a start, still need to add 3rd tier menu, will do soon.

### DIFF
--- a/app/assets/javascripts/select_doctype.js
+++ b/app/assets/javascripts/select_doctype.js
@@ -1,0 +1,81 @@
+// Controls the display of the Document type when creating new assets
+
+Blacklight.onLoad(function() {
+
+	var docTypeJson = "/sample_doctypes.json";
+	var totalDocTypes;
+	//loading json and creating doctype list
+	$.getJSON(docTypeJson).done(function(data) {
+		var $docTypeList = $("#generic_work_document_type_uri");
+		var docVals = [];
+		$("#asset_type_select").change(function() {
+			var $dropdown = $(this);	
+			$docTypeList.html("");	
+			$( "#generic_work_first_document_sub_type_uri" ).remove();
+			$( "#generic_work_second_document_sub_type_uri" ).remove();	
+			$docTypeList.append("<option value=''>Please Select</option>");
+			var key = $dropdown.val();
+			if (key == 'http://definitions.artic.edu/ontology/1.0/type/StillImage') {
+				totalDocTypes = data.asset_types.StillImage.doctypes;
+				}
+			if (key == 'http://definitions.artic.edu/ontology/1.0/type/Text') {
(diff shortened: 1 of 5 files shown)


app/views/records/edit_fields/_document_type_uris.html.erb
```diff
@@ -1,3 +1,8 @@
-<%= f.input key, as: :multi_value_select, collection: DocumentType.options, include_blank: true,
-    selected: f.object.uris_for(:document_type),
-    input_html: { class: 'form-control', multiple: true }, required: f.object.required?(key) %>
+<div class="form-group multi_value required generic_work_document_type_uris managed">
+  	<label class="multi_value required control-label" for="generic_work_document_type_uris">Document type <span class="label label-info required-tag">required</span></label>    
+	
+	<select class="select required form-control form-control" id="docTypes">
```
awead: id should be `generic_work_document_type_uris` and there also needs to be `name="generic_work[document_type_uris][]"` so that the values will be passed in the parameters hash.

app/assets/javascripts/select_doctype.js
```diff
@@ -0,0 +1,55 @@
...
+	//creating subtype list
+	$("#docTypes").change(function() {
+		var $dropdown = $(this);
+		var totalSubtypes = [];
+		$( "#subTypes" ).remove();		
+		var key = $dropdown.val();
+		console.log('key', key);
+		var subTypeVals = [];	
+		if (key) {
+			$("#docTypes").after(" <select class='select required form-control form-control' id='subTypes'><option value=''></option></select>");
```
awead: Because we haven't defined a `subTypes` property, this won't do anything at the moment. Once @scossu has identified the additional predicates, we can change this. For now, though, this is fine. It's just that only the first-level document type will actually get set on the asset.

awead:
Looks good! Just a few minor tweaks to make it work with the form.

app/assets/javascripts/select_doctype.js
scossu: The predicates should be aic:documentType, aic:documentSubType1, aic:documentSubType2.

surreal8:
Hi @awead please check this out. Getting errors on submit, will send separate email in a minute.

app/forms/curation_concerns/generic_work_form.rb
```diff
@@ -5,7 +5,7 @@ class GenericWorkForm < Sufia::Forms::WorkForm
 
     def self.aic_terms
       [
-        :asset_type, :document_type_uri, :first_document_sub_type_uri, :second_document_sub_type_uri,
```
awead, line 8: these need to be there otherwise the form won't submit.

PR #230, opened 2016-07-27
Fix 152

scossu:
@awead need some feedback on this PR.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -68,7 +68,21 @@ class CatalogController < ApplicationController
     config.add_facet_fields_to_solr_request!
 
     # Index view fields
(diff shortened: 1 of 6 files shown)


awead:
Can you fetch the latest develop from upstream and rebase your branch? I've added some improvements that will makes this easier. Also, shouldn't `.label` be `.pref_label`? There's also in a new method in the asset indexer to guard against nil properties.

awead:
You should not have any "merge branch" commits. You could try squashing all these before rebasing.  That will make it much easier.

scossu:
I have been working off the latest develop branch, but I may not be aware of which new methods are available and how they are used. Probably it would be much easier for you to rewrite the commits the right way. I already assigned the ticket to you for the number of assets and department.

PR #231, opened 2016-07-28
issue #450, edits to save work form

surreal8:
Hi Adam, this is almost complete, could you add the visibility options for public and authenticated in the helper files?

### DIFF
--- a/config/locales/sufia.en.yml
+++ b/config/locales/sufia.en.yml
@@ -6,6 +6,8 @@ en:
     institution_name_full:  "The Art Institute of Chicago"
     sort_label:             "Sort the listing of items"
     share_button:           "Upload Your Assets"
+    deposit_agreement:      "LAKE Deposit Agreement"
+    passive_consent_to_agreement: "By clicking 'Save' you agree to the"
     search:
       form:
         option:
@@ -166,3 +168,5 @@ en:
       department:
         text: "Department"
         label_html: "Department"
+      open:
+         label_html: Open Access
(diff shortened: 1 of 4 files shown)


awead:
@surreal8 did you test the form submission with these changes?

surreal8:
@awead yes

awead:
@surreal8 you need to use the config option `Sufia.config.active_deposit_agreement_acceptance` instead. You won't need to change any of the javascript.

config/locales/sufia.en.yml
```diff
@@ -166,3 +168,5 @@ en:
       department:
         text: "Department"
         label_html: "Department"
+      open:
+         label_html: Open Access
```
awead, line 172: something's wrong here. we shouldn't need to add these here, they're already a part of Sufia.

awead:
@surreal8 see https://github.com/projecthydra/sufia/wiki/Customizing-Metadata#creating-a-default-deposit-agreement

config/locales/sufia.en.yml
surreal8, line 172: Default label is "Public" I'm changing that here. Is this not the right place to override the label?

surreal8, line 172: actually these are all overrides of the default values

surreal8:
thanks will take a look now at the agreement

awead:
@surreal8 sorry, I was wrong about the JS. See https://github.com/projecthydra/sufia/issues/2385 but you shouldn't need to override the view html at least.

config/locales/sufia.en.yml
awead, line 172: let me check into this. Open Access shows up other places, so I'm not sure what's going on.

surreal8:
@awead ok, I've updated the pull request leaving the view html with the original code and editing the sufia.rb file

PR #239, opened 2016-08-05
FOR REVIEW ONLY - Rework relationship views.

scossu:
@awead I am not sure about the cause for failing tests, you might have a better grasp.


app/views/curation_concerns/base/show.html.erb
```diff
@@ -17,6 +17,7 @@
   </div>
 </div>
 
+<% if !@presenter.citi_presenter? %>
```
awead, line 20: I would use `unless` here.

app/views/curation_concerns/base/_relationship.html.erb
```diff
@@ -0,0 +1,36 @@
+<h3><% if presenters.first.citi_presenter? %>
```
awead, line 1: I think it would be better to have two kinds of views, one for rel_term and another for rel_term_asset, and have the helper methods call the appropriate one. It avoids have to duplicate this logic. However, if this is doing what you need, we can go with this for now.

scossu, line 1: I agree. This is still manageable so I left it in one file for now. If we want more customization we can split it up.

awead:
This has been added here: 07d161d541eefa9ea7d52e7717b8fbbe70ebe4ab

PR #241, opened 2016-08-10
Parametrize whole URL for Solr and fcrepo, not only port.

scossu:


awead:
The PORT env variables are required. They're used to find available ports during testing.

scossu:
OK, I will rework this. It is just an impediment to me the way I set up my fcrepo (it is behind an Apache vhost).

PR #245, opened 2016-08-12
updating doctypes

surreal8:
using final doc type file and updating javascript to reflect new asset type and json file.
@awead @scossu - we'll need to add a new type in app/views/records/edit_fields/_asset_type.html.erb for 'Other'


scossu:
Sorry, the "Other" column was not in the right place in the reference spreadsheet. See my comment to #658.

PR #275, opened 2016-09-03
Hide edit profile, workaround for #760

scossu:


awead:
@scossu I cherry-picked 21c49966b4977160904d68f02c873bdcabab2124 so we won't have the merge commit. That will make it easier to revert this when the time comes.

PR #299, opened 2016-10-15
adding display behavior to presenter and view

jenlindner:
I think this will fix the problem but suspect it's adding tech debt. Can we discuss, @awead?

### DIFF
--- a/app/services/relationship_presenter_builder.rb
+++ b/app/services/relationship_presenter_builder.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 class RelationshipPresenterBuilder
   attr_reader :citi_uid, :model, :ability, :request
-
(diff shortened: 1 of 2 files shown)


awead:
This looks fine to me, but do we have a test that demonstrates the problem? This is with CITI/Lakeshore integration yes? @scossu can you explain more in the ticket?

scossu:
@awead This is related to ticket #910 on Redmine. The links in the CITI relationship page differed from the ones in the LAKE relationship page, whereas the former lack the resource title in the "add representation" and "add documentation" links. The create page now seems to expect  these fields, and the links in the CITI page laed to a blank page. 

This is a necessary patch for the short term and I agree with @jenlindner that it adds technical debt. A more thorough solution is in ticket #909 where only minimal parameters are passed.

PR #301, opened 2016-10-17
requiring auth for asset link in documents

jenlindner:


awead:
Looks good, but I'll wait to merge until we're sure this solves everything.

PR #312, opened 2016-11-05
Adding Artist and Current Location to Citi Work

jenlindner:
I think this is much better. The key change is in Lakeshore.js, which now gets an array of elements with which to create unique autocompleteControl objects. The autocompleteControl js now has defaults which can be overridden from html data attributes, and the inputmultiselect ruby now supplies them. The defaults might possibly be simpler to provide but this code works and is sufficiently defensive if they aren't provided.

I'd prefer to make the element/jQuery array selector in Lakeshore be configurable also, but this doesn't seem critical right now (unless you disagree, @awead).

The Agent and Place model class property human_readable_type is used in autocompletes, after thinking about it I feel that we would be mixing two use cases to call it from Citi Works for their show view labels. The property is used to identify a class, and should provide an accurate label for itself when it's viewed in its own context. But in the context of a Citi Work, an agent needs to be displayed as an artist. As long as the two use cases exist, it seems to me better to have clear and clearly-named methods for each.

More tests would be better, adding artist.id in the solr_doc model spec as you requested earlier at a minimum. Feedback welcome.


awead:
@jenlindner yes, this look much better, and all looks good. Only one small request: can you separate it into two commits? Have one for the JS/autocomplete config. changes, and a second for the artist and current location stuff. The JS and autocomplete changes aren't required for resolving the original ticket so it'd be nice to have a clear distinction in the commit history. Otherwise, 👍

jenlindner:
Yes, and thanks very much.

jenlindner:
Closing because this PR has been split into two others.

PR #317, opened 2016-11-09
data-attrs on multi inputs can override placeholder endpoint and search length defaults

jenlindner:
@awead would you mind checking this out sometime before Friday? I think it's correct but just don't have strong enough git fu to be certain yet. I rebased against develop after pulling from origin, and the branch now contains dev and the new commit at the top.

I forgot -- you're out til Monday, I think.


awead:
@jenlindner looks good!

PR #321, opened 2016-12-01
delete an asset unless relationships

jenlindner:


scossu:
This seems to be working in test, both for individual and batch deletes. Also batch delete operations involving assets both with and without relationships will delete only assets without relationships. 

Will wait for @awead 's review before resolving ticket.

app/controllers/concerns/aic_asset_delete_behavior.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+module AICAssetDeleteBehavior
+  def delete_asset_and_update_its_collections(curation_concern)
+    curation_concern.in_collection_ids.each do |id|
+      destination_collection = ::Collection.find(id)
+      destination_collection.members.delete(curation_concern)
+      destination_collection.update_index
+    end
```
jenlindner: So @awead, you're saying deleting works from collections they belong to doesn't need to happen?

awead:
@jenlindner yes. If you delete a GenericWork (aka. asset) and it is part of a collection, the collection will be updated to remove that relationship. This happens "automatically" because we're using hydra-pcdm to manage the membership of the collection and the asset. We don't use any of that in the asset to CITI resource relationships.

@awead Ah. Okay, I took that out and refactored a little to clean up. Thanks.

awead:
@jenlindner still on a few things to address here and then I can merge this.

PR #324, opened 2016-12-08
Add general doctypes for Archive, Dataset, Moving Image and Sound.

scossu:


scossu:
@awead Please verify that asset type names match your class names.

PR #328, opened 2016-12-12
Jp2 mogrify

jenlindner:
Hey @awead, this is almost-working code. I'm having trouble overriding what I need to from Hydra-Derivatives, and what I can get to work running the code I can't test. If I follow the structure that's here the tests can find the classes etc, but it's not the best design -- there's an image processor locally, that overrides all of hydra-derivatives' version, because I couldn't get the inheritance to work. I tried to make it work like derivatives/document.rb and derivatives/document_derivatives.rb, and only override the methods I needed to, but just couldn't get it to work. 

And for some reason now (after rebasing, and moving things around to get the test to pass), the output access-jp2 file has zero bytes. 

This is wonky/clunky, but it is a starting place, at least. When it works it resizes jp2 access_masters by running mogrify on them which converts them to jp2s that pass the 'identify' test, and contains a config setting for the local ImageMagick binary to use. Stefano is happy with using the mogrify command on the access-jp2 after it's already been created. It'd be great to get this into develop this week, but I know it's a very rough implementation so far. Any help you can give would be very much appreciated.


awead:
@jenlindner this PR wants to merge into master. I don't think that's what you want. Can you either close it and reopen another, or edit it so that it's merging into develop?

awead:
@jenlindner the issue is resizing images that are wider that 3000 px? Does the code in the PR fix that issue? It would be nice to get a better approach, like you tried with inheritance, but if this is fixing the issue, I'm okay with merging this but it needs a little more documentation in some comments so it's clear what we're doing here.

@awead Thanks for the feedback .. the access files aren't being converted into jp2s in the creation process, and running the mogrify command on them fixes it. The resize goes along well here, and so yes, the code does fix the issues (or will with a minor adjustment, the resize should only run on images that are over 3000px wide), but yeah, I'm disappointed I couldn't get the inheritance together. But if you don't think it's too much debt as is, I'll put in another PR against develop (oops) with a little more documentation.

PR #329, opened 2016-12-13
using mogrify to convert some img derivatives

jenlindner:


config/secrets.yml
```diff
@@ -12,11 +12,12 @@
 
 development:
   secret_key_base: aa56fa52c2e6d296dbabede8f54f61747458742bdf60931917deec5aa66ed9212ce3420fede782d09117c645b911b86c6d017ae3afad6e4620723b717f0fd69a
-
+  hydra_bin_path: "/usr/local/bin/"
```
scossu: These seem an odd fit in `secret.yml`. Is there a better place to put binary configuration?

config/secrets.yml
```diff
@@ -12,11 +12,12 @@
...
 test:
   secret_key_base: 81e7c8558b4bf045e318b4fc5c7527c594357a8b57433c617aa125b5ac48b516677f0591a9f6869dd8542124f79c486ace920481b93040cf4cdff9c21459ef03
-
+  hydra_bin_path: "/usr/local/bin/"
```
scossu: See comment on line 15.

lib/hydra/derivatives/processors/image.rb
```diff
@@ -0,0 +1,92 @@
...
+        output_io.rewind
+
+        str = output_file_service.call(output_io, directives)
+        if directives[:url].include?("access") && directives[:url].include?("jp2")
+          output_file = directives[:url].split("file:")[1]
+          begin
+            _stdin, stdout, _stderr = popen3("#{Rails.application.secrets.hydra_bin_path}identify -format %w #{output_file}")
+            width = stdout.read
+            if width.to_i > 3000
+              Image.execute "#{Rails.application.secrets.hydra_bin_path}mogrify -resize 3000x #{output_file}"
```
scossu: This is still resizing to 3000px on the width. It should resize to 3000px on the longest dimension, i.e. `3000x3000>`

jenlindner:
This should be ready for develop, works with pdfs and psds and tiff files.

scossu (commented):
@jenlindner did you check my two comments on the secrets.yml file? If you and @awead think that is a good fit, I'm OK.

jenlindner:
If we're going to keep private configs in it, it should be in a private repo. But it's a fairly standard place for application configuration, although environment variables would be better. (12 Factor compliant.)

scossu:
My concern is mostly about the fact that files called "secrets" by convention contain highly confidential information and therefore have different location, access, etc. than other configuration files (and our case has not been different so far). 

Is there a separate config file for host names, filesystem paths, etc., or can we create one, to store this kind of information?

jenlindner:
Well, it's not that this file is a bad place for confidential items, it's that it should not be part of the public repo. This file was added in Rails 4.1 for exactly this, secret, confidential configurations. There is even a Capistrano gem to manage syncing this file from private locations when deploying.

Just putting a filesystem path in there while it's in the public repo doesn't seem like a huge security risk to me but you're right, other things definitely would be. 

We used this file from a private repo and the Capistrano gem to manage deploying it at Northwestern. But it wasn't the smoothest process, to be honest, and I think that using environment variables for this would be a better architecture.

https://12factor.net/config

Best,
Jen

awead:
@scossu @jenlindner when you use Capistrano to deploy, you can use its linked files feature to manage this particular file. When the application is deployed, Capistrano will overwrite this file with a symlink pointing to another file on the local server system. What ever is present here in this public file is replaced with a private file that's only available on the server.

jenlindner:
Yes, it is one of the symlinked files.

https://github.com/aic-collections/aicdams-lakeshore/blob/master/config/deploy.rb#L44

And then the capistrano-secrets gem will copy local secrets files to the
server, assuming it's .gitignored, so you end up with a deployment process
that should keep secrets in a private location, symlinked from your app,
and out of the public repo.

But in practice at Northwestern this was cumbersome and error-prone
(keeping everyone's local copy of the secrets file synced was another layer
of work, managing an extra private repo as well). And it doesn't keep
configuration out of the app, which I do feel persuaded is better design.

Jen

jenlindner:
I put LakeshoreTesting.reset_derivatives in a before(:each), clears up data that caused a test to fail when run in a particular order (pdf text extraction after images, seed 14171). And removed what seems to be a no-longer-used second parameter to create_derivatives, no_text: true -- is that okay?

app/jobs/regenerate_asset_derivatives_job.rb
```diff
@@ -7,7 +7,7 @@ def perform(file_set, file_id, filepath = nil)
     return if file_set.video? && !CurationConcerns.config.enable_ffmpeg
     filename = CurationConcerns::WorkingDirectory.find_or_retrieve(file_id, file_set.id, filepath)
 
-    file_set.create_derivatives(filename, no_text: true)
```
awead, line 10: @jenlindner no, this need to be there.

app/models/file_set.rb
```diff
@@ -5,33 +5,27 @@ class FileSet < ActiveFedora::Base
   include Permissions
   self.indexer = FileSetIndexer
 
-  def create_derivatives(filename, opts = {})
-    create_asset_derivatives(filename)
-    create_fulltext_derivatives(filename) unless opts.fetch(:no_text, false)
```
awead, line 10: @jenlindner this as well. we have to not create full text ders. when needed.

spec/models/file_set/file_set_derivatives_spec.rb
```diff
@@ -5,7 +5,7 @@
   let(:file)        { build(:department_file, id: '1234') }
   let(:derivatives) { CurationConcerns::DerivativePath.derivatives_for_reference(file) }
 
-  before { LakeshoreTesting.reset_derivatives }
+  before(:each) { LakeshoreTesting.reset_derivatives }
```
awead: the `:each` is implied and not needed.

spec/models/file_set/file_set_derivatives_spec.rb
```diff
@@ -144,16 +152,5 @@
...
-    context "when specifying no text extraction" do
-      let(:image_file) { File.join(fixture_path, "tardis.png") }
-
-      before { allow(file).to receive(:create_asset_derivatives).with(image_file) }
-
-      it "does not extract full text from the file source" do
-        expect(file).not_to receive(:create_fulltext_derivatives)
-        file.create_derivatives(image_file, no_text: true)
-      end
-    end
```
awead, line 157: we need this test.

app/models/solr_document.rb
```diff
@@ -28,4 +26,12 @@ class SolrDocument
 
   # Do content negotiation for AF models.
   use_extension(Hydra::ContentNegotiation)
+
+  # overrides the MimeTypes included by CurationConcerns::SolrDocumentBehavior in order to add psd types.
+
+  module Hydra::Works::MimeTypes::ClassMethods
```
awead: this doesn't seem like the right place for this. Is this module referenced in the solr document of Sufia or CC?

spec/spec_helper.rb
```diff
@@ -87,3 +87,7 @@
   #   # as the one that triggered the failure.
   #   Kernel.srand config.seed
 end
+
+def in_travis?
```
awead: use LakeshoreTesting.continuous_integration? instead.

app/jobs/regenerate_asset_derivatives_job.rb
awead: @jenlindner ping. see above.

app/models/file_set.rb
awead: @jenlindner ping.

app/models/solr_document.rb
awead: @jenlindner ping.

spec/models/file_set/file_set_derivatives_spec.rb
awead: @jenlindner ping.

spec/models/file_set/file_set_derivatives_spec.rb
awead: @jenlindner ping.

spec/spec_helper.rb
awead: @jenlindner ping.

awead (changes requested):
See comments

spec/models/file_set/file_set_derivatives_spec.rb
jenlindner: I had failing tests without it, with seed 14171, so I added it ("I put LakeshoreTesting.reset_derivatives in a before(:each), clears up data that caused a test to fail when run in a particular order (pdf text extraction after images, seed 14171).")

spec/spec_helper.rb
jenlindner: I can do that.

app/models/solr_document.rb
jenlindner: Finding the right place to override something from the Hydra stack is definitely something I'm wrestling with right now in general. It's included in CC here, in its SolrDocumentBehavior: https://github.com/projecthydra/curation_concerns/blob/613651ef375316d779e4d5f40d608344472cdcb8/app/models/concerns/curation_concerns/solr_document_behavior.rb, which we include above at line 7.

spec/spec_helper.rb
```diff
@@ -87,3 +87,7 @@
   #   # as the one that triggered the failure.
   #   Kernel.srand config.seed
 end
+
+def in_ci_mode?
```
jenlindner: Is this better?

spec/models/file_set/file_set_derivatives_spec.rb
awead: `before` is the same as `before(:each)` so it's likely this might be a transient error.

spec/lib/hydra/derivatives/processors/image_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        expect(mock_image).not_to receive(:resize)
+        expect(mock_image).not_to receive(:format)
+        expect(subject).not_to receive(:write_image)
+        subject.process
+      end
+    end
+  end
+
+  describe "#process and run shell commands" do
+    context "the mogrify command runs", unless: in_ci_mode? do
```
awead: @jenlindner replace with `unless: LakeshoreTesting.continuous_integration?`

awead (changes requested):

spec/spec_helper.rb
awead: @jenlindner see comment below. You don't need an additional method to determine ci.

app/models/solr_document.rb
awead: If you're changing a behavior of an included module, you should override the method or methods that are defined in the module. I'm also not clear on why this is needed. I'm not seeing a corresponding test to solr_document. Could you write a test that demonstrates a failure if this particular bit of code is not present?

spec/models/file_set/file_set_derivatives_spec.rb
jenlindner: I don't think before is the same as before(:each), because the test fails without it because the files are definitely still there from the previous test, with seed 14171 (which runs them in the order in which the persisting files are detected). Adding the :each actually clears the files and the test passes.

app/models/solr_document.rb
jenlindner: I'm overriding a class method in the Hydra Works Mime Types module, which is included in our project by CurationConcerns SolrDocumentBehaviors, so I don't understand where else I should be doing it.

https://github.com/projecthydra/hydra-works/blob/d41be82540eca9d3a76301df3dcea5b74c9d5937/lib/hydra/works/models/concerns/file_set/mime_types.rb#L28

I can write a unit test about this method for the solr document, definitely, but this test relies on the new psd mime type and would fail without it:

https://github.com/aic-collections/aicdams-lakeshore/blob/jp2-mogrify/spec/lib/hydra/derivatives/processors/image_spec.rb#L14-L25

spec/lib/hydra/derivatives/processors/image_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        expect(mock_image).not_to receive(:resize)
+        expect(mock_image).not_to receive(:format)
+        expect(subject).not_to receive(:write_image)
+        subject.process
+      end
+    end
+  end
+
+  describe "#process and run shell commands" do
+    context "the mogrify command runs", unless: LakeshoreTesting.continuous_integration? do
```
jenlindner, line 43: I wondered if this what you had in mind after I changed the spec helper to check this.

spec/models/file_set/file_set_derivatives_spec.rb
jenlindner: No, I'm not right here, :each doesn't solve this problem always. It will error sometimes (for the same reasons, an extra access-pdf leftover) but extremely inconsistently: a seed from running the whole suite with rake spec will *not* cause the error when running it with just the fileset_derivatives_spec.rb. 

I don't know how to fix this yet.

awead: @jenlindner I think this may be related to our intermittent failures. Let me take a look.

awead (approved):

PR #340, opened 2017-01-12
graphicsmagick under the hood except for psd files and imagemagick's convert -flatten

jenlindner:


scossu:
What is the issue with gm and `convert -flatten`?

jenlindner:
It's because gm uses mogrify to do those operations by default, but doesn't
support the -flatten argument. Another way around it is to use gm's convert
for the same thing:

https://github.com/minimagick/minimagick/issues/392

It seems worthwhile and I started working on a test for that today too.

Jen

app/models/file_set.rb
```diff
@@ -17,9 +17,16 @@ def create_derivatives(filename, opts = {})
     # pdf for text assets
     # second thumbnail for CITI
     def create_asset_derivatives(filename)
+      # byebug
```
awead: remove.

app/models/file_set.rb
```diff
@@ -17,9 +17,16 @@ def create_derivatives(filename, opts = {})
...
       case mime_type
       when *self.class.image_mime_types
-        Hydra::Derivatives::ImageDerivatives.create(filename, outputs: image_outputs)
+        if mime_type =~ /psd/i || mime_type =~ /vnd.adobe.photoshop/i
+          MiniMagick.with_cli(:imagemagick) do
+            Hydra::Derivatives::ImageDerivatives.create(filename, outputs: image_outputs)
+          end
+        else
+          Hydra::Derivatives::ImageDerivatives.create(filename, outputs: image_outputs)
+        end
```
awead, line 24: why not:
``` ruby
when *image_mime_types_with_cli
  MiniMagick.with_cli(:imagemagick) do
    Hydra::Derivatives::ImageDerivatives.create(filename, outputs: image_outputs)
  end
when *self.class.image_mime_types
  Hydra::Derivatives::ImageDerivatives.create(filename, outputs: image_outputs)
[...]

def image_mime_types_with_cli
  ["psd", "vnd.adobe.photoshop"]
end
```
This makes the method less branchy. Alternatively, you could define `image_mime_types_with_cli` in config/initializers/hydra_works.rb if you need to use these mime types outside of this class.

spec/models/file_set/file_set_derivatives_spec.rb
```diff
@@ -46,6 +47,19 @@
         ]
       end
 
+      context "with a psd mime type" do
+        before {
```
awead: use do/end with multilines. {} are reserved for one-liners.

awead (changes requested):

app/models/file_set.rb
jenlindner: Yes, cleaner.

spec/models/file_set/file_set_derivatives_spec.rb
jenlindner: Right.

awead (approved):

PR #342, opened 2017-01-19
adding vertical wrapper to simple_form setup

jenlindner:
hi @awead, I found that adding the wrapper to the simple_form setup block solves the bug (1379 in redmine, the wrapper couldn't be found and the form relying on it didn't load, file_set edit page couldn't render). Not sure how to test though in a view though, or if we need a test for this?

### DIFF
--- a/config/initializers/simple_form.rb
+++ b/config/initializers/simple_form.rb
@@ -1,15 +1,23 @@
 # frozen_string_literal: true
 
-# Use this setup block to configure all options available in SimpleForm.
 SimpleForm.setup do |config|
   # Wrappers are used by the form builder to generate a
   # complete input. You can remove any component from the
   # wrapper, change the order or even add your own to the
   # stack. The options given below are used to wrap the
   # whole input.
-  config.wrappers :default, class: :input,
-                            hint_class: :field_with_hint,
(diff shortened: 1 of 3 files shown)


spec/views/curation_concerns/file_sets/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  end
+
+  before do
+    allow(controller).to receive(:current_user).and_return(user)
+
+    allow(AICUser).to receive(:find_by_nick).with(user).and_return(user)
+
+    view.lookup_context.prefixes.push 'curation_concerns/base'
+    render template: 'curation_concerns/file_sets/edit.html.erb', locals: { curation_concern: file_set }
```
jenlindner: I can't really work through how to stub the file set such that the form thinks it has what it needs, and I can just check that it renders with the all of the UI elements it's given (including the wrapper). I looked at sufia's fileset form view tests for ideas. The above gives me a Blacklight record not found error. Explicitly trying to pass the form stub an array with main app as a double and the file_set like this ``` allow(form).to receive([main_app, file_set]) ``` gives me a different error, a blacklight one.

PR #347, opened 2017-01-29
handle unauthorized errors with more info when uploading duplicate asset

jenlindner:
Hi @awead, this is more like what we discussed yesterday, it uses a presenter class (yay!) and no longer rescues the error, so the controller and controller tests are more simple. I decided to just handle generic_work authorization for now, although @scossu is also interested in having an unauthorized page for all situations and I could continue with that next week. The presenter class could use refactoring and there could be more assertions in the tests, but I wanted to get this out this afternoon. 

Also I'm suddenly having fedora problems running the entire test suite so couldn't check to see if everything passes, but I know the new ones for this code do.

### DIFF
--- a/app/controllers/curation_concerns/generic_works_controller.rb
+++ b/app/controllers/curation_concerns/generic_works_controller.rb
@@ -16,6 +16,15 @@ class CurationConcerns::GenericWorksController < ApplicationController
     end
     report_status_and_redirect(asset_with_relationships, batch: false)
   end
+  # we have to override this action in Curation Concerns application controller behavior class
+
+  def deny_access
+    presenter = UnauthorizedPresenter.new(params[:id])
+    render template: '/error/unauthorized',
+           formats: [:html],
+           status: 401,
+           locals: { presenter: presenter }
+  end
 
   protected
 
--- a/app/presenters/unauthorized_presenter.rb
+++ b/app/presenters/unauthorized_presenter.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+class UnauthorizedPresenter
+  def initialize(id)
+    @asset = GenericWork.find(id)
+  end
+
+  def depositor_last_name
+    @asset.aic_depositor.family_name.nil? ? "" : @asset.aic_depositor.family_name
+  end
+
+  def depositor_first_name
+    @asset.aic_depositor.given_name.nil? ? "" : @asset.aic_depositor.given_name
+  end
+
+  def depositor
+    @depositor = depositor_last_name.empty? && depositor_first_name.empty? ? "citi_support@artic.edu" : "#{depositor_first_name} #{depositor_last_name}"
+  end
+
+  def title
+    @asset.to_solr["title_tesim"].first
+  end
+
+  def thumbnail
+    @asset.to_solr["thumbnail_path_ss"]
+  end
+
+  def message
+    "You are not authorized to see this asset. Please contact #{depositor} if you would like to request access to it."
+  end
+end
--- a/app/views/error/unauthorized.html.erb
+++ b/app/views/error/unauthorized.html.erb
@@ -1,0 +1,15 @@
+<div class="alert alert-danger">
+  <%= presenter.message %>
+</div>
+<div class="row">
+  <div class="col-xs-12">
+    <div class="col-xs-3">
+      <%= image_tag("#{presenter.thumbnail}") %>
+    </div>
+    <div class="col-xs-3">
+      <p class="unauthorized_asset_title">
+        <%= presenter.title %>
+      <p>
+    </div>
+  </div>
+</div>
--- a/app/assets/stylesheets/aic_styles.scss
+++ b/app/assets/stylesheets/aic_styles.scss
@@ -182,3 +182,7 @@ div.home_share_work {
 table.relationships img {
 	max-height: 150px;
 }
+
+p.unauthorized_asset_title {
+  margin-top: 25%;
+}
--- a/app/services/duplicate_upload_verification_service.rb
+++ b/app/services/duplicate_upload_verification_service.rb
@@ -8,7 +8,7 @@ class DuplicateUploadVerificationService
   end
 
   def duplicates
-    existing_filesets
+    existing_filesets.empty? ? [] : [GenericWork.find(existing_filesets.first.parent.id)]
   end
 
   private
--- a/spec/controllers/curation_concerns/generic_works_controller_spec.rb
+++ b/spec/controllers/curation_concerns/generic_works_controller_spec.rb
@@ -38,6 +38,7 @@ describe CurationConcerns::GenericWorksController do
       expect(response).to be_unauthorized
     end
   end
+
   describe "#destroy" do
     context "with relationships" do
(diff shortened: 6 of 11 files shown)


awead:
@jenlindner good. this is much better. Yes, you'll need to override `deny_access`, but note that in CurationConcerns, it is expecting an argument:

https://github.com/projecthydra/curation_concerns/blob/master/app/controllers/concerns/curation_concerns/application_controller_behavior.rb#L23

app/controllers/curation_concerns/generic_works_controller.rb
```diff
@@ -16,6 +16,16 @@ def destroy
...
   end
+  # we have to override this action in Curation Concerns application controller behavior class
+
+  def deny_access
+    locals = UnauthorizedPresenter.new(params).render_asset_401
+
+    render template: '/error/unauthorized',
+           formats: [:html],
+           status: 401,
+           locals: { title: locals[:title], message: locals[:message], depositor_first_name: locals[:depositor_first_name], depositor_last_name: locals[:depositor_last_name], thumbnail: locals[:thumbnail] }
```
awead: why all the local variables? The only thing you need to pass to the view is the presenter. The presenter should have methods for title, message, depositor, etc.

app/controllers/sufia/uploads_controller.rb
```diff
@@ -73,6 +73,11 @@ def asset_attributes
         params.fetch(:generic_work, nil) || params.fetch(:batch_upload_item, nil)
       end
 
+      def duplicate_generic_work(fileset)
```
awead: can you refactor DuplicateUploadVerificationService to return the parent GenericWork instead of the FileSet?

awead: That way `@duplicate_upload` is the actual asset that contains the duplicate file set and not just the file set itself.

app/controllers/curation_concerns/generic_works_controller.rb
awead: the above should look something like:
``` ruby
presenter = UnauthorizedPresenter.new(params[:id])
render template: '/error/unauthorized',
  formats: [:html],
  status: 401,
  locals: { presenter: presenter }
 ```

File: app/presenters/unauthorized_presenter.rb [hunk 3]
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+class UnauthorizedPresenter
+  def initialize(params)
+    @params = params
```
awead: instead of capturing the params, build your GenericWork here:
``` ruby

class UnauthorizedPresenter
  attr_reader :asset
  def initialize(id)
    @asset = GenericWork.find(id) if GenericWork.exists?(id)
  end
```

awead: note, the work may not exist, so `@asset` could be nil here.

app/presenters/unauthorized_presenter.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def get_depositor(depositor)
+    depositor_last_name = depositor.family_name.nil? ? "" : depositor.family_name
+    depositor_first_name = depositor.given_name.nil? ? "" : depositor.given_name
+    @depositor = depositor_last_name.empty? && depositor_first_name.empty? ? "citi_support@artic.edu" : "#{depositor_first_name} #{depositor_last_name}"
```
awead, line 16: make this a method:
``` ruby
def depositor
  depositor_last_name = asset.aic_depositor.family_name.nil? ? "" : asset.aic_depositor.family_name 
  [etc...]
end
```

app/presenters/unauthorized_presenter.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def message
+    "You are not authorized to see this asset. Please contact #{@depositor} if you would like to request access to it."
+  end
+
+  def render_asset_401
+    generic_work = GenericWork.find(@params[:id])
+    get_depositor(generic_work.aic_depositor)
+    @gw_solr_doc = generic_work.to_solr
+    @title = generic_work.to_solr["title_tesim"].first
```
awead: same here, make `title` a method.

app/presenters/unauthorized_presenter.rb
awead: I take that back... you can assume the asset exists because it contains a duplicate file set.

app/presenters/unauthorized_presenter.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    @thumbnail = @gw_solr_doc["thumbnail_path_ss"]
```
awead: make `thumbnail` a method...

app/views/error/unauthorized.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="alert alert-danger">
+  <%= message %>
```
awead: now for the fun part... just call presenter for everything.. `presenter.message`

app/views/error/unauthorized.html.erb
```diff
@@ -0,0 +1,15 @@
...
+</div>
+<div class="row">
+  <div class="col-xs-12">
+    <div class="col-xs-3">
+      <%= image_tag("#{thumbnail}") %>
```
awead: presenter.thumbnail

app/views/error/unauthorized.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    </div>
+    <div class="col-xs-3">
+      <p class="unauthorized_asset_title">
+        <%= title %>
```
awead: presenter.title

awead (changes requested):

scossu:
Note: this unauthorized page should be specific to GenericWork but generic otherwise, i.e. not only for duplicates. I am not sure if some of @awead 's comments apply to this scenario.

app/presenters/unauthorized_presenter.rb
scossu: @awead See my comment below: https://github.com/aic-collections/aicdams-lakeshore/pull/347#issuecomment-275953096 this could be any unauthorized page for GenericWork. Not sure if I'm in the right context here though...

app/controllers/sufia/uploads_controller.rb
jenlindner: I'm having trouble with one of the sufia uploads controller tests for it, working on that now. The rest of the changes, though, much better, thanks.

jenlindner:
This has the duplicates service returning an asset, and the tests are passing (I just ran rspec spec/ folder by folder locally, Travis will find anything I missed), so I think it might be ready. (I didn't add an argument to deny_access, seems to work without it? And I don't seem to need one.)

@scossu, yes that's in line with what @awead outlined for going forward, this page doesn't contain copy regarding duplicates, and the presenter that gets it information will be refactored to be more generic, usable sitewide.

awead:
@jenlindner this looks great. thanks!  I have some minor tweaks for the presenter, if you want, otherwise, you can merge this.

jenlindner:
Thanks! And thanks very much for all the help @awead. I'm going to merge it and will want to to get the tweaks into the presenter next.

PR #348, opened 2017-02-03
Request access

jenlindner:
Hi @awead, this branch isn't quite ready for a merge but I would like you to review a little, if you wouldn't mind. I think it's structured pretty well overall and the test coverage should be good too, I made some effort. And I'm hoping you can help me with one test, for the request_access_presenter - for some reason I just can't get the stub for GenericWorkObject to be called in it, and it errors.


app/mailers/request_access_mailer.rb
```diff
@@ -0,0 +1,10 @@
+class RequestAccessMailer < ActionMailer::Base
+
+  def request_access(presenter)
+    @depositor_first_name = presenter.depositor_first_name
+    @requester_pretty_name = presenter.requester_pretty_name
+    @resource_title = presenter.title
+    @resource_id = presenter.resource_id
```
awead: I would omit these 4 variables and simply set `@presenter`

app/presenters/request_access_presenter.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+class RequestAccessPresenter < UnauthorizedPresenter
```
awead, line 2: insert `attr_reader :requester` here

app/presenters/request_access_presenter.rb
```diff
@@ -0,0 +1,26 @@
...
+  def initialize(resource_id, user)
+    super(resource_id)
+    @requester = user
+  end
+
+  def requester
+    @requester
+  end
```
awead: remove this method

app/presenters/request_access_presenter.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def requester_email
+    "#{@requester}" + "@artic.edu"
```
awead: you don't need the "+" : `"#{requester}@artic.edu"`

app/presenters/request_access_presenter.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def depositor_email
+    "#{@resource.aic_depositor.nick}" + "@artic.edu"
```
awead: same here. Note that because `requester` is an attr_reader, we can omit the class variable designation "@"

app/presenters/request_access_presenter.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def requester_pretty_name
+    aic_user = AICUser.find_by_nick(@requester)
```
awead: simplify to `#{aic_user.family_name} #{aic_user.given_name}"` You could also do: `[aic_user.family_name, aic_user.given_name].join(" ")` which would allow for cases where either might be nil.

app/presenters/unauthorized_presenter.rb
```diff
@@ -1,30 +1,34 @@
 # frozen_string_literal: true
 class UnauthorizedPresenter
```
awead, line 2: use another attr_reader for resource here: `attr_reader :resource`

app/presenters/unauthorized_presenter.rb
```diff
@@ -1,30 +1,34 @@
...
   end
 
   def depositor_first_name
-    @asset.aic_depositor.given_name.nil? ? "" : @asset.aic_depositor.given_name
+    @resource.aic_depositor.given_name.nil? ? "" : @resource.aic_depositor.given_name
   end
 
   def depositor
-    @depositor = depositor_last_name.empty? && depositor_first_name.empty? ? "citi_support@artic.edu" : "#{depositor_first_name} #{depositor_last_name}"
+    @depositor = depositor_last_name.empty? && depositor_first_name.empty? ? "#{@resource.aic_depositor.nick}" : "#{depositor_first_name} #{depositor_last_name}"
```
awead: I think these three methods could be simplified to just one:
``` ruby 
def depositor
  return resource.aic_depositor.nick unless resource.aic_depositor.family_name && resource.aic_depositor.given_name
  "#{resource.aic_depositor.given_name} #{resource.aic_depositor.family_name}"
end
```

awead: Note, depositor doesn't need to be a class variable.

app/presenters/unauthorized_presenter.rb
```diff
@@ -1,30 +1,34 @@
 # frozen_string_literal: true
 class UnauthorizedPresenter
   def initialize(id)
-    @asset = GenericWork.find(id)
+    @resource = GenericWork.find(id)
```
awead: Consider creating a SolrDocument here and then you could use methods for all the terms below.
`@resource = SolrDocument.new(GenericWork.find(id).to_solr)`
then delegate all your methods to resource:
`delegate :title, :family_name, :given_name, :thumbnail, to: :resource`
see https://github.com/aic-collections/aicdams-lakeshore/blob/master/app/models/concerns/solr_document_extensions.rb#L68-L92

app/presenters/unauthorized_presenter.rb
```diff
@@ -1,30 +1,34 @@
...
+    @resource.to_solr["thumbnail_path_ss"]
+  end
+
+  def resource_id
+    @resource.id
   end
 
   def message
-    "You are not authorized to see this asset. Please contact #{depositor} if you would like to request access to it."
+    "You are not authorized to see this asset."
```
awead: I'd move this to the view as an i18n string.

app/views/request_access_mailer/request_access.html.erb
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <p>Dear <%= @depositor_first_name %>,</p>
```
awead: presenter.dep_first_name and so on...

app/views/request_access_mailer/request_access.text.erb
```diff
@@ -0,0 +1 @@
+hi
```
awead: more here?

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+describe ApplicationController do
+  routes { Rails.application.routes }
+  include_context "authenticated saml user"
+
+  def setup_stubs
```
awead: make this a `before do` block and it will run each time.

spec/mailers/request_access_mailer.erb_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  it 'renders the sender email' do
+    mock_presenter
+
+    RequestAccessMailer.request_access(presenter).deliver_now
+
+    expect(ActionMailer::Base.deliveries.first.from).to eq(["user1@artic.edu"])
+  end
+
+  after(:each) do
```
awead: each is always implied.

spec/mailers/request_access_mailer.erb_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  allow(presenter).to receive(:requester_email).and_return("user1@artic.edu")
+  allow(presenter).to receive(:title).and_return("Sample")
+  allow(presenter).to receive(:resource_id).and_return("1234")
+  allow(presenter).to receive(:depositor_email).and_return("kcole@artic.edu")
+end
+
+describe RequestAccessMailer, :type => :mailer do
+  let(:presenter) { double }
+
+  before(:each) do
```
awead: each is implied.

spec/mailers/request_access_mailer.erb_spec.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+def mock_presenter
```
awead: move this into your `before do` block below. Also add `RequestAccessMailer.request_access(presenter).deliver_now` because you're calling it each time as well.

awead (changes requested):
Looks good! My suggestions are to get some of the finer points and use Ruby's idioms to your advantage.

awead (approved):

jenlindner:
hi @awead, one last thing. I put the domain into an environment variable for mailer urls, is that okay with you?

PR #349, opened 2017-02-08
adding create lab work button to asset show view

jenlindner:
@awead, is it okay to add imaging_uid to the work_presenter terms, so that it delegates to solr, to use in a view? I'd expect this to be just fine but had to change to some work model tests to adjust, just wanted to be sure I did that correctly. I *think* I changed the tests to say the model shouldn't respond to this term because the presenter delegates it to solr but I'm not sure about that.


awead:
@jenlindner yep that looks good!

jenlindner:
thanks! 😸

PR #352, opened 2017-02-12
Imaging uid fix

awead:
@jenlindner there was a bug with your recent commit to add the lab button. You weren't accounting for non-assets, and it was throwing an error because non-asset presenters don't have imaging_uid delegated to them. As far as I can tell, the lab button should only render for assets and not any non-assets.

I fixed it up and refactored it. Take a look and see if this is what you're after.

@scossu this bug is in test right now, so it'll affect testing until it's resolved.


app/views/curation_concerns/base/_show_actions.html.erb
```diff
@@ -6,7 +6,7 @@
       <%= link_to "Edit", edit_polymorphic_path([main_app, presenter]), class: 'btn btn-default' %>
       <%= link_to "Delete", [main_app, presenter], class: 'btn btn-danger', data: { confirm: "Delete this #{presenter.human_readable_type}?" }, method: :delete if presenter.deleteable? %>
   <% end %>
-  <% unless presenter.imaging_uid.empty? %>
+  <% if presenter.imaging_uid.present? %>
```
jenlindner, line 9: Yeah that's a lot better.

spec/views/curation_concerns/base/_show_actions.html.erb_spec.rb
```diff
@@ -2,33 +2,41 @@
...
 
 describe 'curation_concerns/base/_show_actions.html.erb' do
-  let(:asset) { build(:asset, id: '1234', pref_label: 'FineArt') }
-  let(:solr_doc) { SolrDocument.new(asset.to_solr) }
-  let(:presenter) { AssetPresenter.new(solr_doc, ability) }
   let(:ability) { double }
-  let(:page) { Capybara::Node::Simple.new(rendered) }
+  let(:page)    { Capybara::Node::Simple.new(rendered) }
+
+  before do
```
jenlindner, line 8: Yeah it's cleaner to do them all up here.

jenlindner:
@awead. Thanks. This is better, and makes more sense why I was having to subtract and add the term in the Work Presenter tests. Going to merge now.

PR #356, opened 2017-03-09
Force change for documents types set to nil

awead:
Because we're using a Definition class that is not AF::Base, we get peculiar errors, such as when you set a Definition property to nil, the change is not recognized in the AF stack. Here we are forcing the change notification so we can successfully remove any document type that has been set.


awead:
@jenlindner can you test your PR against this branch and see if it fixes your problem? You shouldn't need to do anything to the actors, just passing an empty string in the params hash should be sufficient to trigger the removal of a document type.

jenlindner:
Beautimus. Thank you! And I learned more too :-)

PR #357, opened 2017-03-20
fixes for subtypes and second subtypes

jenlindner:
Hi @awead -- this still needs a little cleanup (some contexts in the feature tests need it blocks or to be commented out, for example), but I wanted to get it out there and reviewed by @scossu and at least checked out by you now. Not the best js in the word but it is fixing the bugs.

### DIFF
--- a/app/assets/javascripts/select_doctype.js
+++ b/app/assets/javascripts/select_doctype.js
@@ -55,139 +55,295 @@
 
             documentSubType: function(model, term, option) {
                 var html =
+=======
+    var select_doctype = {
+      //todo - this would be nicer as a case statement with a regex pattern to match against, instead of repeating the url parts each time.
+      getAssetType: function(data, assetType) {
+        var totalDocTypes;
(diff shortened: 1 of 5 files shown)


jenlindner:
Ugh, @awead, we're getting errors in the feature tests from travis that I just don't get locally, and one even from a test that shouldn't have been affected. Should I try to recreate them locally, or do you have a silver bullet? Might be that I'm not using Capybara helper methods where I ought to be (though I thought I was), and therefore causing timing problems.

awead:
@jenlindner try running it locally with the same seed. There could be an issue with stray assets in Solr and Fedora because tests are randomized. I would also try to run as many tests within one page load.

jenlindner:
I decided a new strategy is a good idea. Now I'm only testing the user flow that had exposed the bugs the js (and back end persistence) changes fix in one new feature spec. But still seeing test problems with Travis that I don't get locally, with the same seeds. 

I'll keep thinking about what might be causing this and what to do about it.

jenlindner:
And now I don't know *what* to think about the build passing.

PR #361, opened 2017-03-23
Js doctypes - one last test.

jenlindner:


jenlindner:
Closing because commit history not clean enough.

PR #365, opened 2017-04-12
adding resque-cleaner gem

jenlindner:
Hi @awead, would you mind merging this before you deploy tonight? It's just a gem and a one-line config file, no code changes. The cleaner gem adds some helpful visibility from the resque admin, and we needed to regenerate the derivatives in test today, so it would be nice to have this visibility on that big job. 

Thanks, @jenlindner


awead:
Looks great, but why all the failures?

PR #369, opened 2017-04-27
riif server and universal viewer

jenlindner:
@awead, this PR is not something that I want to get into the next release, but soon after would be good. Stefano asked for a layout change yesterday that solved some presentation issues the viewer had had with pdfs (in the smaller size allotted for them, pdfs overflowed vertically over the bottom of the viewer), and this branch might be closer to an acceptable merge than I'd thought. It needs some manifest service tests and a test for the display_image presenter and probably more, but, would you mind taking a look at the general structure to let me know what you think? 

Also, when the code migrates to Hyrax this will quite likely come out, but it's a very nice to have presentation until then.

Thanks.


awead:
@jenlindner this is too big! It looks like you've cloned the entire universalviewer repo into public. Can you use a submodule instead? Or either put all the UV public/bower_includes in a different commit.

jenlindner:
What, 20 million files is a lot??

Okay :-)

PR #370, opened 2017-05-03
Universal viewer riiif

jenlindner:
hi @awead, the universal-viewer files are in a separate commit in this branch.


awead:
@jenlindner thanks. So, I don't think including the entire repo in `public/` is going to work. For reference, lakeshore's line count is 119,289. This commit alone is over twice that, which, for your github stats is totally awesome! Congrats! But seriously, it's going to bloat the repo way too much. It would be much more prudent to use a submodule here:
```
cd public/bower_includes
git submodule add https://github.com/UniversalViewer/universalviewer
```
This way it's kept separate from the lakeshore repo, but all the files are still there. Let me know if you have any questions about this.

jenlindner:
@awead, I pretty much took this structure straight from hyku: https://github.com/projecthydra-labs/hyku/tree/master/public/bower_includes/universalviewer/dist/uv-2.0.1. Is this your only criticism so far of this PR?

Actually there's a bug, I removed some code in models/fileset that we need. 

I'll work on this more tomorrow.

awead:
@jenlindner that's some kind of release. It's not the entire repo. So far, yes, I have more to say about the other commit. It's mostly fine, though.

jenlindner:
@awead, ah, I get what you mean now. Okay glad it's mostly fine with you, I'll see what kind of shape I can get it into by the end of the day.

PR #373, opened 2017-05-05
Universal Viewer Dist and RIIIF and IIIF Manifest files

jenlindner:


scossu:
Just to be clear, this is not a candidate for merging into 1.1.0. There is still some testing to be done.

awead:
Note. Don't merge this until after 1.1.0 is released.

app/assets/stylesheets/viewer.scss
```diff
@@ -0,0 +1,19 @@
+.viewer {
+  height: 500px;
+  padding: 10px;
+  iframe {
+    // position: absolute;
```
awead: can you remove the iframe if this is commented-out?

app/controllers/curation_concerns/generic_works_controller.rb
```diff
@@ -4,8 +4,46 @@ class CurationConcerns::GenericWorksController < ApplicationController
...
+  def manifest
+    headers['Access-Control-Allow-Origin'] = '*'
+    respond_to do |format|
+      format.json do
+        render json:
+          ManifestService.new(presenter).manifest_builder
+      end
+    end
+  end
+  # overriding CC show action to render the html template for assets that uses the layout specific to universal viewer which occupies all columns, no sidebar
```
awead: This comment should be moved down so that it's right about `def show`. Also, could you explain what specific line or lines in the method you're overriding? I'm wondering that if you're changing the template or view code, you could change the template or just override the view code instead of the controller code.

Gemfile
```diff
@@ -21,6 +22,10 @@ gem 'sass-rails', '~> 5.0'
 gem 'sdoc', '~> 0.4.0', group: :doc
 gem 'turbolinks'
 gem 'uglifier', '~> 3.0'
+gem 'openseadragon', '0.3.0'
```
awead: openseadragon is already included in line 17

app/models/file_set.rb
```diff
@@ -75,6 +75,10 @@ def image_mime_types_with_cli
       ["psd", "vnd.adobe.photoshop"]
     end
 
+    def image_mime_types
```
awead: We shouldn't need this. It's in `config/initializers/hydra_works.rb`

app/models/solr_document.rb
```diff
@@ -26,4 +26,12 @@ class SolrDocument
 
   # Do content negotiation for AF models.
   use_extension(Hydra::ContentNegotiation)
+
+  # overrides the MimeTypes included by CurationConcerns::SolrDocumentBehavior in order to add psd types.
+
+  module Hydra::Works::MimeTypes::ClassMethods
```
awead: Hmm. I'm wondering why this is needed here. It's been overridden in `config/initializers/hydra_works.rb`

app/presenters/aic_work_show_presenter.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+class AICWorkShowPresenter < Sufia::WorkShowPresenter
+  def manifest_url
+    manifest_helper.polymorphic_url([:manifest, self])
+  end
+
+  def description
+    'a brief yet clear description'
```
awead: This seems like a good candidate for an entry in the locales file!

app/presenters/aic_work_show_presenter.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+class AICWorkShowPresenter < Sufia::WorkShowPresenter
```
awead: You've defined a new class here but it only has one subclass, so why the inheritance?

app/presenters/asset_presenter.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
-class AssetPresenter < Sufia::WorkShowPresenter
+# AICWorkShowPresenter subclasses Sufia::WorkShowPresenter
+# and adds tiling image and canvas displaying behaviors
+class AssetPresenter < AICWorkShowPresenter
```
awead: AssetPresenter is the only class inheriting from AICWorkShowPresenter, so I would merge the two classes and put the methods in AICWorkShowPresenter in this class.

app/presenters/concerns/displays_image.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+# This gets mixed into FileSetPresenter in order to create
+# a canvas on a IIIF manifest
+module DisplaysImage
+  extend ActiveSupport::Concern
+
+  def display_image
+    return nil unless FileSet.exists?(id)
```
awead: This is hitting Fedora and we shouldn't do that in a presenter context. Run the solr query here and reutrn nil if the query is nil.

app/presenters/concerns/displays_image.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    url = DerivativePath.access_path(id)
+    doc = ActiveFedora::SolrService.get("id:#{id}")
+    height = doc["response"]["docs"][0]["height_is"]
+    width = doc["response"]["docs"][0]["width_is"]
```
awead: You build a SolrDocument here:
``` ruby
hit = ActiveFedora::SolrService.query("id:#{id}")
doc = SolrDocument.new(hit)
```
Now, you can call `doc.height` and `doc.width`
see: https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/solr_service.rb#L47

app/renderers/universal_viewer.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+class UniversalViewer
+  include Singleton
```
awead, line 3: Is UniversalViewer supposed to be a singleton? Why? I've heard singletons are "bad" just curious...

app/services/iiif_authorization_service.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+class IIIFAuthorizationService
```
awead: Could put some class-level documentation in here? I'm guessing this is something that `Riif` uses? A service that takes a controller as a param caught me off-guard.

app/views/curation_concerns/base/show_asset.html.erb
```diff
@@ -0,0 +1,38 @@
+<% provide :page_title, @presenter.page_title %>
```
awead, line 1: Could add some comments to indicate what line(s) are being overridden?

config/secrets.yml
```diff
@@ -17,4 +17,4 @@ test:
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
 production:
-  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
```
awead: I would revert this change. It isn't germane to the PR, and the key should be indented.

app/services/manifest_service.rb
```diff
@@ -0,0 +1,68 @@
...
+    @presenter.solr_document["hasRelatedImage_ssim"].first
+  end
+
+  def media_is_pdf?
+    FileSet.find(file_set_id).to_solr["file_format_sim"].include?("pdf (Portable Document Format)")
+  end
+
+  def manifest_builder
+    # we need a ManifestFactory object and to be able to add media_sequences to it; its to_h is connected to a lot of internal state, is not just the creation of a hash
+    builder = IIIFManifest::ManifestFactory.new(@presenter)
```
awead, line 59: I think the better approach here is to create your own factory:
``` ruby
class PdfManifestFactory < IIIFManifest::ManifestFactory
  def add_media_sequences(manifest, media_info)
    [...]
end
```
But, isn't `manifest` above just `self.to_h`?

spec/spec_helper.rb
```diff
@@ -87,3 +87,7 @@
   #   # as the one that triggered the failure.
   #   Kernel.srand config.seed
 end
+
+def in_travis?
```
awead: use https://github.com/aic-collections/aicdams-lakeshore/blob/master/spec/support/lakeshore_testing.rb#L31 instead

awead (changes requested):
@jenlindner looks good overall. It needs a rebase off of develop first, and then let's talk about the remaining bits.

app/presenters/aic_work_show_presenter.rb
jenlindner: @awead a brief yet clear entry! working on it now :-)

app/views/curation_concerns/base/show_asset.html.erb
jenlindner: They're added, at line 15.

app/services/manifest_service.rb
jenlindner: This is the main aspect I couldn't quite get, I tried in a variety of ways to make this work with its own factory. I don't think this is the worst way for the time being. And to_h is a method in the factory that adds an awful lot of behavior, isn't just the to_hash one might expect.

app/renderers/universal_viewer.rb
jenlindner: This is another thing I took pretty much without any questions from Hyku, so couldn't really answer. We could ask Trey or Justin if you'd like, though.

jenlindner:
hi @awead, I have a question. There's caching happening on the rails side that prevents the viewer from loading a good portion of the time. I've experimented some this afternoon with different types of headers to force expiration with no luck, but a little more research turned up turbolinks as a likely culprit too. Does that sound plausible to you? I started looking into ways to leave them enabled for everything except the controller and action in question, but thought I'd ask for your feedback as well. Thanks.

awead:
@jenlindner yes, that seems very likely. We could disable it altogether, if needed, but maybe there's a subtler way. Have you asked @jcoyne or others involved with Hyku?

jenlindner:
Okay, I tested with turbolinks disabled site-wide and the problem disappears. Yay to finding the root cause.

So then, as to how to disable them only in this view: one way is to use a data attribute in link tags, but our code doesn't use an html link tag to load the viewer, and the viewer generates its own html. Another, which I think might be the way to go, is to put "data-no-turbolink" into the body tag, which I think might best be accomplished in a layout. (I'm getting information on disabling them from this post: http://blog.flightswithfriends.com/post/53943440505/how-to-disable-turbolinks-in-rails-4)

We use the Curation Concerns layout code on the show pages: 

https://github.com/projecthydra/curation_concerns/blob/master/app/controllers/concerns/curation_concerns/curation_concern_controller.rb#L9-L10

https://github.com/projecthydra/curation_concerns/blob/master/app/controllers/concerns/curation_concerns/themed_layout_controller.rb#L13-L21

This creates the body tag: https://github.com/projecthydra/curation_concerns/blob/master/app/views/layouts/curation_concerns.html.erb#L1

So, one option is to override this layout in LAKE to add the attribute to the body tag.

Another option might be to override the method with_themed_layout(view_name = nil) in LAKE to take a parameter for adding an attribute to the body tag. 

Do you have any recommendations, @awead @jcoyne? I haven't tested this yet, also, don't know for certain that the attribute in the body tag would solve the problem.

awead:
@jenlindner I've seen `data-no-turbolink` before as an option. Does that fix the problem? If not, I'd say post to hydra-tech about this and see what Hyku is doing, otherwise I think we'd just have to nix turbolinks.

jenlindner:
That does fix the problem, thanks @awead. I added it to a new layout, added the layout to the generic works controller and added a controller test that checks for both the new template and the layout in the show action.

awead (approved):

PR #374, opened 2017-05-16
Resque cleaner

jenlindner:
This branch mainly contains a gem that allows for resque job management and a config change that has resque jobs log to the resque log, for easier debugging/control during ingests, derivatives, etc.


jenlindner:
hi @awead, this doesn't quite have the test I wanted, but it would be checking that logging goes to the resque log, not confirming app code changes (there aren't any). please let me know if you think the branch is okay.

PR #375, opened 2017-05-16
adding resque cleaner and more logging

jenlindner:
hi @awead. This branch enables the cleaner tab in the resque admin, and contains a config change that has resque jobs log to the resque log, for easier debugging/control during ingests, derivatives, etc. I wanted to test it well but can't quite test the logging today, but that's not an app code change, so maybe can wait ....

### DIFF
--- a/app/jobs/create_derivatives_job.rb
+++ b/app/jobs/create_derivatives_job.rb
@@ -10,7 +10,7 @@ class CreateDerivativesJob < ActiveJob::Base
     filename = CurationConcerns::WorkingDirectory.find_or_retrieve(file_id, file_set.id, filepath)
 
     file_set.create_derivatives(filename)
-
+    Resque.logger.info("queue running fine: #{CurationConcerns.config.ingest_queue_name}")
     # Reload from Fedora and reindex for thumbnail and extracted text
     file_set.reload
     file_set.update_index
--- a/Rakefile
+++ b/Rakefile
@@ -3,9 +3,9 @@
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
 require File.expand_path('../config/application', __FILE__)
+require 'resque/tasks'
 
 Rails.application.load_tasks
-
 unless Rails.env.production?
   load File.expand_path('../tasks/dev.rake', __FILE__)
   load File.expand_path('../tasks/test.rake', __FILE__)
--- a/config/initializers/resque_admin.rb
+++ b/config/initializers/resque_admin.rb
@@ -4,6 +4,5 @@ class ResqueAdmin
     # TODO: restrict to a group (see issue #4)
     current_user = request.env['warden'].user
     !current_user.blank?
-    # current_user.groups.include? 'umg/up.dlt.scholarsphere-admin'
   end
 end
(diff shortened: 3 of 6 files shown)


app/jobs/create_derivatives_job.rb
```diff
@@ -6,11 +6,12 @@ class CreateDerivativesJob < ActiveJob::Base
   # @param [String] file_id identifier for a Hydra::PCDM::File
   # @param [String, NilClass] filepath the cached file within the CurationConcerns.config.working_path
   def perform(file_set, file_id, filepath = nil)
+    # raise StandardError
```
awead: Remove?

app/jobs/lakeshore/test_fail_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class Lakeshore::TestFailJob < ActiveJob::Base
```
awead: Is this ever used anywhere else or is it just to test the logger? If so, you can move the job definition to the spec test itself:

``` ruby
describe "A failing job" do
  class TestFailJob < ActiveJob::Base
    [etc]
  end
  
  it "raises an error" do
    expect { TestJob.perform_now }.to raise_error(StandardError)
  end
end
```

config/deploy.rb
```diff
@@ -44,6 +44,9 @@
   'config/property_permissions.yml',
   'config/redis.yml',
   'config/resque-pool.yml',
+  'config/resque-web.rb',
+  'config/initializers/resque_config.rb',
+  'config/initializers/resque_admin.rb',
```
awead: What's the rationale behind having Capistrano manage these? We usually do this if there's sensitive information in them, or if the configuration needs to change between servers. Is that the case here?

awead (changes requested):

awead:
@jenlindner looks fine. I just had a few questions.

config/deploy.rb
jenlindner: Well, because these files aren't going to change that much, and ought to be preserved through deploys of app code changes.

app/jobs/lakeshore/test_fail_job.rb
jenlindner: Yeah, the testing that goes along with this isn't okay yet, but it's testing a a pretty benign configuration. I do want to spend more time with it and have some sort of test that confirms the new configuration that logs to a different log actually logs to it, but I won't have the time to do that for a few more days, and there's a need for it before then. I could remove these tests since they're not really doing much, until I have them in shape that will support the application.

config/deploy.rb
awead: OK. I guess it doesn't seem like a good choice to me because that's more files to have to manage on the server side. If they don't change that much, then I don't see an incentive to manage them elsewhere. If you deploy the application to a new server, you now have three more files that you have to manage.

jenlindner:
@awead I went with your requests. If this seems fine to you, would you mind deploying to production as soon as you get a chance? I said I'd do it today but I'm not feeling well, and it would be the first time I've done it, which hasn't worked yet, and I don't think I'm up to the troubleshooting now.

awead:
@jenlindner will do! Hope you feel better.

awead (approved):

PR #389, opened 2017-09-06
Add Art Purchase Funding Plan and Loan Appeal Letter doctypes.

scossu:


scossu:
@awead Travis is failing with a mention to departments. I am not sure what's wrong in the commit that would affect those tests.

awead:
@scossu this is a problem with the test suite. We get transient failures now and then. I've restarted the build and it should pass now.

PR #404, opened 2017-11-03
tweaking prefix for local env, LCL is much better on the eyes than LO…

RudyOnRails:
…C (LOC may infer location)


RudyOnRails:
Did this ever go green? I assume so since you merged, but maybe you found
that the red builds were not related to my diff! Travis being Travis...

awead:
@RudyOnRails no, there's the unrelated issue with Jasmine. I have another PR that separates the Jasmine build so we can see what's actually failing, but for now I think we can ignore it on Travis and make sure we run it locally.

PR #415, opened 2017-11-21
move dev:prep data creation to seeds.rb from factory_girl

RudyOnRails:
LAKE DEV citi data does not exist and dev:prep uses FactoryGirl gem which isn't loaded when "production" so moving to seeds.

Will add other citi resources at some point after I'm done testing something in DEV which started this whole snowball :)


awead:
@RudyOnRails I was thinking put all the resources that are created in dev:prep in here. But, we might have to check that it doesn't try to created them if they already exist.

RudyOnRails:
closing until it's worked on again.

PR #429, opened 2018-01-04
2338: Replace "preferred representation" in edit form with a "star" r…

RudyOnRails:
…adio button

* https://cits.artic.edu/redmine/issues/2338
* updated JS to include new elements when adding a rep
* added stylings for stars
* reordered nonasset’s representations array so that pref rep is [0]
* removed pref_rep_uri from rep_terms and corrected spec
* created new subclass for HiddenMultiSelectStarsInput and moved star HTML and logic over
* removed no-longer-needed _preferred_representation_uri.html.erb partial
* updated value for simple_form :as to use new subclass HiddenMultiSelectStarsInput
* added in thead, tr, and th’s for column titles


RudyOnRails:
closing in favor of https://github.com/aic-collections/aicdams-lakeshore/pull/431

PR #430, opened 2018-01-06
2339: Replace "preferred representation" in view page with star icon

RudyOnRails:
* https://cits.artic.edu/redmine/issues/2339

* remove preferred_reps table partial from nonasset show view
* tweak render_asset_relationship helper method to accept optional 3rd arg and pass it to asset_relationship partial as a local
* implement star_or_not method inside _asset_relationship partial & helper
* passing new 3rd arg of pref_rep_id to render partial method so that reps table can compare to each rep id to star it or not
* clicking on stars SHOULD NOT do anything on show action, increase specificity of element that jquery binds event to
* reorder citiwork presenter rep presenters that get passed to partial so starred is displayed as first <tr>
* work edit - document tr's should not have 'preferred' <th> and <td> when adding new ones

### DIFF
--- a/app/assets/javascripts/lakeshore.js
+++ b/app/assets/javascripts/lakeshore.js
@@ -82,7 +82,7 @@ $(function() {
         parent_table.prepend(current_row);
     }
 
-    $("tbody").on('click', '.fa-star-o', function(){
+    $("table.representation_uris").on('click', '.fa-star-o', function(){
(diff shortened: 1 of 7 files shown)


RudyOnRails:
closing in favor of https://github.com/aic-collections/aicdams-lakeshore/pull/431

PR #445, opened 2018-01-31
Bug #1686: Improve duplicate checking on Lakeshore UI ingestion

RudyOnRails:
# https://cits.artic.edu/redmine/issues/1686

## QUESTIONS
- [x] if we just add the postgres db location to the DuplicateUploadVerificationService will it cause problems? Yes b/c if the job checks the PG table for the checksum and it exists, the jobs won't perform.
- [x] Where will we calculate checksums on new files? We should probably calculate the checksum on each of the uploaded_files inside some model.
- [x] Where will we save the checksum string? We will save it in the `uploaded_files` table, under a new column called `checksum`.

## SUPPORTING INFO/FACTS
* currently, DuplicateUploadVerificationService is used in the following places: `CreateAssetJob`, `Sufia::UploadsController`, and `Lakeshore::IngestController`. The Job should check for a checksum in solr, and the UI and API controllers should check both Solr and the DB
*  inside the batch_uploads controller (possibly a callback inside a AR model), calculate checksums on each uploaded_file and store in Postgres.
* split into separate methods, calling the `.solr` class method from the Jobs, and calling the `.solr_and_db` from the controllers
* calculate checksum as soon as UploadsController receives successful upload, and store in tmp so when batch is submitted, those checksums can be stored in DB, per Stefano, this way the web process won't take forever to calculate checksums on large batches and respond to the web client

## Acceptance Criteria:
- [x] When a user uploads a file that has already has begun ingestion (but *not* indexed in Solr with a checksum), they should receive a warning just like when a file has already been indexed.
- [x] When a user uploads a file that does not exist in Solr, the status of the file should be "submitted"
- [x] When an uploaded file is finally indexed in Solr with a checksum, that checksum should be cleared from the DB table.

## TO DO:
- [x] add a new column, of type :[string](https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L76), to the uploaded_files table
- [x] on XHR upload, add a [callback](http://guides.rubyonrails.org/v4.0/active_record_callbacks.html#available-callbacks) to store checksum as [Digest::SHA1](http://ruby-doc.org/stdlib-2.4.0/libdoc/digest/rdoc/Digest/SHA1.html) the same string that ends up in Fedora, and Solr


Gemfile.lock
```diff
@@ -214,6 +214,7 @@ GEM
       tins (~> 1.6)
```
RudyOnRails: no gem changes needed, this was extra from trying to figure out if I could do `receive_message_chain` w/ `.with` in the latest rspec

need to checkout this file and repush

app/actors/delete_uploaded_file_actor.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
```
RudyOnRails: don't need this actor anymore - pls check it out and then stash/shelve it locally.

app/controllers/sufia/batch_uploads_controller.rb
```diff
@@ -21,4 +23,9 @@ def build_form
       @form = form_class.new(curation_concern, current_ability)
       @form.parameterized_relationships = ParameterizedRelationships.new(params)
     end
+
+    def set_uploaded_file_status
+      uploaded_file_ids = params["uploaded_files"]
+      Sufia::UploadedFile.flip_status(uploaded_file_ids)
```
RudyOnRails: the name of this class method is a little vague - does it flip it on its side? upside down? 😄 
rename to something like `change_status` and perhaps pass the desired status string:
`Sufia::UploadedFile.change_status(file_ids, "begun_ingestion")`

app/controllers/sufia/uploads_controller.rb
```diff
@@ -33,7 +33,7 @@ def validate_asset_type
 
       def validate_duplicate_upload
         return if duplicate_upload.empty?
-        render json: { files: [duplicate_error_message] }
+        render json: { files: [duplicate_error_message, begun_ingestion] }
```
RudyOnRails: this looks like temp code, make sure to wrap up!

app/controllers/sufia/uploads_controller.rb
```diff
@@ -67,6 +67,14 @@ def duplicate_error_message
...
       end
 
+      def begun_ingestion
+        {
+            error:          t('lakeshore.upload.errors.begun_ingestion', name: uploaded_file.original_filename),
+            duplicate_path: polymorphic_path(duplicate_upload.first),
+            duplicate_name: duplicate_upload.first.to_s
+        }
+      end
+
```
RudyOnRails: consider having one private method that returns a hash for when the checksum is in solr, and a different hash for when it has begun ingestion.

app/jobs/concerns/adding_files_to_works.rb
```diff
@@ -13,7 +13,7 @@ def perform(work, uploaded_files)
       end
 
       attach_content(actor, uploaded_file.file)
-      uploaded_file.update(file_set_uri: file_set.uri)
+      uploaded_file.update_attribute(:file_set_uri, file_set.uri)
```
RudyOnRails: Is it good to skip validations here? It is a valid parameter, but what if file_set is nil?

I remember changing to `update_attribute` so that validations were skipped, but perhaps we customize the validation to only validate checksum :on => :create ?

RudyOnRails: if file_set is nil, we have a bigger problem 💥

app/jobs/create_derivatives_job.rb
```diff
@@ -2,6 +2,15 @@
 class CreateDerivativesJob < ActiveJob::Base
   queue_as CurationConcerns.config.ingest_queue_name
 
+  after_perform do |job|
+    fileset = job.arguments.first
+    fileset_uri = fileset.uri.to_s
+    uploaded_file = Sufia::UploadedFile.find_by_file_set_uri(fileset_uri)
+    unless FileSet.where(digest_ssim: uploaded_file.checksum).empty?
+      uploaded_file.update_attribute(:status, "ingested")
```
RudyOnRails: if we make the checksum validation only validate on :create, we can change this to `.update`

app/models/concerns/uploaded_file.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+module UploadedFile
+  include Lakeshore::ChecksumService
+  extend ActiveSupport::Concern
+
+  included do
+    extend ClassMethods
+    after_initialize :set_status, if: "status.nil?"
+    before_validation :calculate_checksum
+    validates :checksum, presence: true, checksum: true
```
RudyOnRails: we should probably only validate `on: :create`

app/models/concerns/uploaded_file.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def set_status
+      self.status = "new"
+    end
+
+  module ClassMethods
+    def flip_status(uploaded_files_ids)
+      uploaded_files_ids.each do |uploaded_file_id|
+        find(uploaded_file_id).update_attribute(:status, "begun_ingestion")
```
RudyOnRails: @awead do you think I should use .update_attribute on the backend where validation isn't that important b/c we have control over the attributes? Or should we always be super safe and spark validations with .update?

RudyOnRails:
unit builds (PR and PUSH) were failing. grabbed travis log and restarted. here are failures so far:
```
[31m1535 examples, 15 failures, 6 pending[0m

Failed examples:

[31mrspec ./spec/controllers/sufia/uploads_controller_spec.rb:72[0m [36m# Sufia::UploadsController updating the use uri use_uri should eq #<RDF::Vocabulary::Term:0x41ff1a4 URI:http://definitions.artic.edu/ontology/1.0/type/IntermediateFileSet>[0m
[31mrspec ./spec/jobs/attach_files_to_work_job_spec.rb:15[0m [36m# AttachFilesToWorkJob with each of the different use files attaches the files to the work and updates them[0m
[31mrspec ./spec/actors/replace_file_actor_spec.rb:20[0m [36m# ReplaceFileActor an intermediate file when it is present replaces the content of the existing file set[0m
[31mrspec ./spec/presenters/asset_presenter_spec.rb:101[0m [36m# AssetPresenter#publishable when false publishable should equal "No"[0m
[31mrspec ./spec/presenters/asset_presenter_spec.rb:106[0m [36m# AssetPresenter#publishable when true publishable should equal "Yes"[0m
[31mrspec ./spec/presenters/asset_presenter_spec.rb:94[0m [36m# AssetPresenter#public_domain when true public_domain should equal "Yes"[0m
[31mrspec ./spec/presenters/asset_presenter_spec.rb:89[0m [36m# AssetPresenter#public_domain by default public_domain should equal "No"[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller_spec.rb:263[0m [36m# Lakeshore::IngestController#update with no duplicates with an original file when one already exists replaces the file[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller_spec.rb:252[0m [36m# Lakeshore::IngestController#update with no duplicates with an original file when none exists title [0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller_spec.rb:280[0m [36m# Lakeshore::IngestController#update with no duplicates with a preservation master file when none exists title [0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller_spec.rb:291[0m [36m# Lakeshore::IngestController#update with no duplicates with a preservation master file when one already exists replaces the file[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller_spec.rb:183[0m [36m# Lakeshore::IngestController#update with no duplicates with an intermediate file when one does not exist title [0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller_spec.rb:209[0m [36m# Lakeshore::IngestController#update with no duplicates with an intermediate file when one already exists when the depositor is another user in the asset depositor's department changes the file_set's depositor[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller_spec.rb:319[0m [36m# Lakeshore::IngestController#update with no duplicates with a legacy file when one already exists replaces the file[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller_spec.rb:308[0m [36m# Lakeshore::IngestController#update with no duplicates with a legacy file when none exists title [0m
```

Will run locally. Travis failing and tests passing locally is nothing new, it's just been a pain and I'd like to document this for when we throw time at it and resolve it.

RudyOnRails:
I figured a [spreadsheet](https://docs.google.com/spreadsheets/d/1D_s2tOED-at4z7fgtXaONfCSEKMsvHj2OX96RQXpYPk/edit?usp=sharing) was a better tool to gain some insight into this muddle.

.rspec-local
```diff
@@ -1 +0,0 @@
---format documentation
```
RudyOnRails: can you get rid of this commit/change? perhaps do a git checkout of it from develop branch.

app/controllers/curation_concerns/generic_works_controller.rb
```diff
@@ -9,6 +9,8 @@ class CurationConcerns::GenericWorksController < ApplicationController
   self.show_presenter = AssetPresenter
   skip_load_and_authorize_resource only: :manifest
 
+  after_action :set_uploaded_file_status, only: [:create]
```
RudyOnRails: oh nice, so when individual assets are created by imaging, they will also set the uploaded files to "begun_ingestion"...

app/controllers/curation_concerns/generic_works_controller.rb
```diff
@@ -60,4 +62,9 @@ def build_form
       @form.current_ability = current_ability
       @form.action_name = action_name
     end
+
+    def set_uploaded_file_status
+      uploaded_file_ids = params["uploaded_files"]
```
RudyOnRails: what happens if there are no uploaded_files when submitting an external file?

app/jobs/concerns/adding_files_to_works.rb
RudyOnRails: After further thought, I don't think this has anything to do with the main issue, so if it's important, just commit separately or cherry pick it later.

app/models/batch_upload.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+class BatchUpload < ActiveRecord::Base
+  has_many :uploaded_files
```
RudyOnRails: You may not even use this, as I don't recall any lines like `@batch_upload.uploaded_files`, can probably omit, although it doesn't hurt anything. plus we could always use it in rails console for troubleshooting, so keep it 👍

app/validators/checksum_validator.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+class ChecksumValidator < ActiveModel::Validator
+  include Rails.application.routes.url_helpers
+  def validate(record)
+    @record = record
+    @uploaded_file = @record.file
+    @duvs_obj = DuplicateUploadVerificationService.new(@uploaded_file)
+    if !@duvs_obj.duplicates.empty?
+      record.errors.add(:checksum, in_solr_error_message)
+    elsif Sufia::UploadedFile.begun_ingestion.map(&:checksum).include?(@record.checksum)
```
RudyOnRails: this looks like it can be refactored to be less taxing. I would just chain a where onto the begun_ingestion scope and see if is empty or not.

app/validators/checksum_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+      record.errors.add(:checksum, begun_ingestion_error_message)
+    elsif Sufia::UploadedFile.where(batch_upload_id: @record.batch_upload_id).map(&:checksum).include?(@record.checksum)
```
RudyOnRails: We can probably split these 3 validations up into separate methods.

app/validators/checksum_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+        error:          I18n.t('lakeshore.upload.errors.begun_ingestion', name: @record.file.filename)
+      }
+    end
+
+    def already_in_batch_error_message
+      {
+        error:          I18n.t('lakeshore.upload.errors.already_in_batch', name: @record.file.filename)
+      }
+    end
+end
```
RudyOnRails: these 3 methods look like they can be combined into a dynamic one.

app/services/duplicate_upload_verification_service.rb
```diff
@@ -16,14 +24,13 @@ def duplicates
...
-  # @return [Array<FileSet>]
-  def duplicate_file_sets
-    return [] if file.nil?
-    FileSet.where(digest_ssim: fedora_shasum)
-  end
-
   private
 
+    # returns array of one parent asset
+    def parent_asset
```
awead: @RudyOnRails I don't see this method getting called anywhere.

app/models/batch_upload.rb
awead: @RudyOnRails what about calling this `UploadedBatch` or even `UploadedFileBatch`? It might help distinguish it from the other classes that deal with batch edits, etc. like `BatchEditItem` and `BatchUpload`.

app/jobs/create_asset_job.rb
```diff
@@ -2,10 +2,14 @@
 class CreateAssetJob < CreateWorkJob
   before_perform do |job|
     job.arguments[2].fetch(:uploaded_files, []).each do |id|
-      file = Sufia::UploadedFile.find(id)
-      unless DuplicateUploadVerificationService.unique?(file)
+      @sufia_uploaded_file = Sufia::UploadedFile.find(id)
```
awead: why the @ variable?

app/services/duplicate_upload_verification_service.rb
RudyOnRails: Good catch.

app/jobs/create_asset_job.rb
RudyOnRails: Because I need to use the variable on the instance later in the `after_perform` down below. Does that make sense?

app/models/batch_upload.rb
RudyOnRails: @awead hmm, I don't see any other Ruby classes named `BatchUpload`. Don't we want to keep it similar to what `Sufia::BatchUploadsController` handles?

RudyOnRails:
Sweet, so I remove a method that is not used anywhere, not even in tests, and the builds fail. I love this stuff! ❤️ woohoo! 

```
Finished in 20 minutes 3 seconds (files took 8.34 seconds to load)
1542 examples, 1 failure, 6 pending
Failed examples:
rspec ./spec/jobs/batch_asset_create_job_spec.rb:37 # BatchAssetCreateJob#perform with uploaded files creates two assets
```

passes on my machine:
```
~/RubymineProjects/aicdams-lakeshore (1686_improve_dupe_checks) $ rspec ./spec/jobs/batch_asset_create_job_spec.rb:37
Run options: include {:focus=>true, :locations=>{"./spec/jobs/batch_asset_create_job_spec.rb"=>[37]}}

Randomized with seed 14814

BatchAssetCreateJob
  #perform
    with uploaded files
      creates two assets

Top 1 slowest examples (0.62505 seconds, 14.5% of total time):
  BatchAssetCreateJob#perform with uploaded files creates two assets
    0.62505 seconds ./spec/jobs/batch_asset_create_job_spec.rb:37

Finished in 4.31 seconds (files took 7.81 seconds to load)
1 example, 0 failures

Randomized with seed 14814

Coverage report generated for RSpec to /Users/kmusio/RubymineProjects/aicdams-lakeshore/coverage. 1605 / 2226 LOC (72.1%) covered.
```

I'm "this" close to changing the config.before to restore before _every_ example, not just features, and dealing with day long builds.

RudyOnRails:
Yeah, Travis ~doesn't~ don't got time for that:
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

app/jobs/create_asset_job.rb
awead: @RudyOnRails Yes. I'm not certain that instance variables carry over across before and after blocks. You might want to double-check that, unless the tests are already proving they do.

app/models/batch_upload.rb
awead: @RudyOnRails I was thinking about `BatchUploadItem`, which might makes things confusing later to other developers, or even to us 6 months from now when we've forgotten all about this. The `BatchUploadItem` is actually the model that interacts with `BatchUploadForm` and `BatchUploadController`. I was suggesting `UploadedBatch` because it aligns more with `UploadedFile`. The batch here is a grouping of uploaded files. There's also no metadata either, and the concept of a batch upload in Sufia is uploading multiple assets, with both files, relationships, and metadata.

Anyway, It's only a suggestion. Feel free to come up with other names too!

app/models/concerns/uploaded_file.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+module UploadedFile
```
awead, line 2: @RudyOnRails How is this different than `Sufia::UploadedFile`? I would think you'd be overriding that class instead of creating a new one.

app/services/checksum_service.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+module ChecksumService
```
awead, line 2: @RudyOnRails You should include this module in `DuplicateUploadVerificationService` and remove the duplicate code. It doesn't matter if the methods are private or not.

app/validators/checksum_validator.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+class ChecksumValidator < ActiveModel::Validator
+  include Rails.application.routes.url_helpers
+  def validate(record)
+    @record = record
+    @uploaded_file = @record.file
```
awead: @RudyOnRails I don't see this instance var. used anywhere else.

app/validators/checksum_validator.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+class ChecksumValidator < ActiveModel::Validator
+  include Rails.application.routes.url_helpers
+  def validate(record)
+    @record = record
```
awead: @RudyOnRails I would make this an `attr_reader` and just use `record` throughout for clarity. It seems more idiomatic to me.

app/validators/checksum_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+    @uploaded_file = @record.file
+    @duvs_obj = DuplicateUploadVerificationService.new(@uploaded_file)
```
awead: @RudyOnRails you could memoize this in a private method for a cleaner look:
``` ruby
def duvs_object
  @duvs_object ||= DuplicateUploadVerificationService.new(@record.file)
end
```
Also, why not `verification_service` for the method/variable? It's a little more wordy, but would show up if you were searching for the words.

app/validators/checksum_validator.rb
awead: @RudyOnRails yeah, probably, but I think the readability here is good.

app/validators/checksum_validator.rb
awead: @RudyOnRails yes. It's hard to see the logic elsifs because the chain methods are very long. What would it look like if you could rewrite it as a case statement? Could you create an internal class the encapsulated all the where logic and simply reported the type of error?

spec/controllers/sufia/uploads_controller_spec.rb
```diff
@@ -6,38 +6,23 @@
   include_context "authenticated saml user"
 
   let(:duplicates) { [] }
-  before { allow(controller).to receive(:duplicate_upload).and_return(duplicates) }
+  # before { allow(controller).to receive(:duplicate_upload).and_return(duplicates) }
```
awead: @RudyOnRails remove?

app/models/concerns/uploaded_file.rb
RudyOnRails: It isn't, they are the same - I include this module in the other: https://github.com/aic-collections/aicdams-lakeshore/pull/445/files#diff-b1fe55db50c712fef0673345e5b9c0d9R43
```
+    config.to_prepare do
+      Sufia::UploadedFile.include UploadedFile
+    end
```

app/services/checksum_service.rb
RudyOnRails: I did at first, but then some tests were blowing up, in the DuplicateUploadVerificationService specs because the test objects were failing validation, IIRC.

app/validators/checksum_validator.rb
RudyOnRails: I used it in:
`@duvs_obj = DuplicateUploadVerificationService.new(@uploaded_file)`

app/validators/checksum_validator.rb
RudyOnRails: I could do that. I'm often torn between using ivar's and readers/writers. I think a few weeks ago I  decided it was best to use `attr_*`'s for when messages need to be sent to the instance from outside the class.

app/validators/checksum_validator.rb
RudyOnRails: Great points, I like these suggestions.

spec/controllers/sufia/uploads_controller_spec.rb
RudyOnRails: Forgot about this one, thanks 👍

RudyOnRails:
TODO:

- [x] double check that ivars carry over from `before_perform` to `after_perform`
- [x] swap out `BatchUpload` with `UploadedBatch` and see if it all jives
- [x] remove commented line from spec/controllers/sufia/uploads_controller_spec.rb
- [x] move @duvs_obj to [memoized helper](https://github.com/aic-collections/aicdams-lakeshore/pull/445#discussion_r184901042) 
- [x] make `@record` an attr_reader
- [x] obviously squash this beast

PR #456, opened 2018-03-14
2647: Update API does not respect depositor parameter

RudyOnRails:
# REDMINE URL: https://cits.artic.edu/redmine/issues/2647


RudyOnRails:
This issue is resolved and tests pass. They are rerunning now after I rebased and cleaned up my commits, but no reason they should fail, @scossu's testing of intermediate was 👍 too. depositor is applied to only the fileset.

New bug related to non-intermediate filesets causing nils has been opened so that we can break down unrelated issues into smaller tickets. new bug issue is https://cits.artic.edu/redmine/issues/2773

anyone disagree with merging this?

PR #462, opened 2018-04-05
2827: Set asset as preferred representation in ingest API

RudyOnRails:
https://cits.artic.edu/redmine/issues/2827

branched off develop and reverted the revert. Bringing in 1794 (well, the _old_ 1794, that was deleted, and is now 2827), again.


app/controllers/lakeshore/ingest_controller.rb
```diff
@@ -48,7 +48,7 @@ def validate_duplicate_upload
       end
 
       def validate_preferred_representations
-        return unless represented_resources.present?
+        return if force_preferred_representation? || represented_resources.empty?
```
RudyOnRails, line 51: @awead Could we combine this logic into a skip_before_filter, then run only if these conditions are met, reducing the responsibility of this method? I get a bit confused that the methods get called, but don't run. But that could just be a design pattern I am unfamiliar with. I feel it should be swapped, and only called if certain logic is true, but my thinking could not be the best.

spec/models/lakeshore/ingest_spec.rb
```diff
@@ -205,4 +205,33 @@
...
+      it { is_expected.to be(false) }
+    end
+
+    context "when set to true" do
+      let(:params) { { metadata: { force_preferred_representation: "true" } } }
+      it { is_expected.to be(true) }
+    end
+
+    context "when set to 1" do
+      let(:params) { { metadata: { force_preferred_representation: 1 } } }
```
RudyOnRails: @awead I believe all params will be wrapped in strings...so you may want to change to `"1"` and fix implementation. Not sure how often we will use `"1"` vs `"true"`, but this example will live in it's own world, and any requests with "1" would not force the preferred representation.

app/models/lakeshore/ingest.rb
```diff
@@ -57,13 +57,19 @@ def represented_resources
       end
     end
 
+    def force_preferred_representation?
+      return true if force_preferred_representation == "true" || force_preferred_representation == 1
+      false
```
RudyOnRails: @awead Perhaps we could take out the "1" case all together, and just keep in line with true/false? similar to `duplicate_check` flag?

`force_preferred_representation == "true"`

app/models/lakeshore/ingest.rb
```diff
@@ -57,13 +57,19 @@ def represented_resources
...
+    end
+
     private
 
       def register_terms(metadata)
         return unless metadata.present?
         @document_type_uri = metadata.fetch(:document_type_uri, nil)
         @preferred_representation_for = metadata.fetch(:preferred_representation_for, [])
         @ingestor = find_or_create_user(metadata.fetch(:depositor, nil))
+        @force_preferred_representation = metadata.fetch(:force_preferred_representation, false)
```
RudyOnRails: @awead Just curious, any reason why we use instance variables instead of `attr_writer`/`accessor`'s? I don't think it makes a difference, unless the writer were custom.

spec/controllers/lakeshore/ingest_controller/create_spec.rb
```diff
@@ -0,0 +1,218 @@
+# frozen_string_literal: false
+require 'rails_helper'
+
+describe Lakeshore::IngestController do
```
RudyOnRails: @awead Why split up the spec into 2? Getting too big?? Do we lose any DRY'ness? If we keep it split, all I ask is that we add wrapping describe blocks so in rspec's documentation format, we can differentiate between `#create` and `#update`.

app/models/lakeshore/ingest.rb
RudyOnRails: @awead Should we keep request flags/conditions out of the metadata? - keep them top level like duplicate_check?
```
-F 'force_preferred_representation=true'\
```

app/controllers/lakeshore/ingest_controller.rb
awead: @RudyOnRails I find the guard clauses keep the logic contained in the method, instead of separating logic in the filter and then in the action in the method. It also keeps the pattern consistent. Each before filter uses a guard clause. Since the logic varies per method, you'd need to make separate `skip_before_filter` declarations for each method, each with an if clause and an only clause. To me, that increases the cognitive load of each of those lines, requiring the reader to keep more in their head until they get the actual method call. By using just `before_filter`, the only responsibility of that line is to call the method for specific actions. This reduces the complexity of that one line. Then, the logic regarding what the actual method does is relegated to the method itself, and not split between the method and its filter invocation.

spec/models/lakeshore/ingest_spec.rb
awead: @RudyOnRails During a meeting with one of the citi devs awhile back, there was a question about this. IIRC, sometimes they use 0 or 1 for their booleans. I wasn't sure, so I was leaving it open-ended. @scossu does this matter? Otherwise, we could only accept a "true" string for the value of True.

app/models/lakeshore/ingest.rb
awead: @RudyOnRails double-check with @scossu, but that would be the simpler route, if that's acceptable to the clients using the API.

app/models/lakeshore/ingest.rb
awead: @RudyOnRails these are write-once variables, so a full attr_writer isn't needed.

RE: metdata, yes, I had the same thought. Dunno. It makes sense to keep it out of metadata, but check with @scossu to see what the client's preferences/expectations are.

spec/controllers/lakeshore/ingest_controller/create_spec.rb
awead: @RudyOnRails Yes, it was getting too big and the indentation, given all the difference contexts, was becoming difficult to read. Although these are controller tests, they read (and really are) more like feature tests, and need a little more flexibility to their description options. I'm pushing up a solution.

app/controllers/lakeshore/ingest_controller.rb
RudyOnRails: I see your point. We would need to add two more `skip_before_action`'s each with a (new & separate) guard clause method, just to make the validation guts simpler.

Maybe one day all this model validation logic (`before_filter`s and 4+ methods) will be in the model, and we can call `.save`'s on the `Lakeshore::Ingest` object, firing off model validations, adhering to skinny-controllers/fat-models.

app/models/lakeshore/ingest.rb
RudyOnRails: Will do - he's out today but I will send him a summary email to get his opinion.

spec/models/lakeshore/ingest_spec.rb
RudyOnRails: Leaving 0 or 1 is totally doable if the client chooses to communicate that way; my point is that the parameter checked inside the model is checking for `1` when the param clients will pass in will probably be converted to a string of `"1"`, since this is not an ActiveRecord attribute that is typecasted back to an integer.

RudyOnRails:
Restarting the two failing unit builds (one PR, one PUSH), although they should be identical as this branch is rebased on top of develop.

Travis has been giving false negatives. Not sure if it's my fault, but if you restart the build, they often go green...

```
Finished in 18 minutes 1 second (files took 8.93 seconds to load)
[31m1542 examples, 12 failures, 6 pending[0m

Failed examples:

[31mrspec ./spec/jobs/attach_files_to_work_job_spec.rb:15[0m [36m# AttachFilesToWorkJob with each of the different use files attaches the files to the work and updates them[0m
[31mrspec ./spec/jobs/attach_files_to_work_job_spec.rb:25[0m [36m# AttachFilesToWorkJob with no use specified attaches the files to the work and updates them[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller/update_spec.rb:71[0m [36m# Lakeshore::IngestController replacing files with an original file when none exists title [0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller/update_spec.rb:82[0m [36m# Lakeshore::IngestController replacing files with an original file when one already exists replaces the file[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller/update_spec.rb:138[0m [36m# Lakeshore::IngestController replacing files with a legacy file when one already exists replaces the file[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller/update_spec.rb:127[0m [36m# Lakeshore::IngestController replacing files with a legacy file when none exists title [0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller/update_spec.rb:99[0m [36m# Lakeshore::IngestController replacing files with a preservation master file when none exists title [0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller/update_spec.rb:110[0m [36m# Lakeshore::IngestController replacing files with a preservation master file when one already exists replaces the file[0m
[31mrspec ./spec/controllers/lakeshore/ingest_controller/update_spec.rb:45[0m [36m# Lakeshore::IngestController replacing files with an intermediate file when none exists title [0m
[31mrspec ./spec/jobs/batch_asset_create_job_spec.rb:86[0m [36m# BatchAssetCreateJob#perform with an external file and no label creates an asset with a url[0m
[31mrspec ./spec/jobs/batch_asset_create_job_spec.rb:37[0m [36m# BatchAssetCreateJob#perform with uploaded files creates two assets[0m
[31mrspec ./spec/jobs/batch_asset_create_job_spec.rb:61[0m [36m# BatchAssetCreateJob#perform with an external file and label creates an asset with a url[0m
```

RudyOnRails:
2 failures on PUSH unit build, going to restart build.
```
Finished in 19 minutes 10 seconds (files took 8.41 seconds to load)
1542 examples, 2 failures, 6 pending
Failed examples:
rspec ./spec/jobs/attach_files_to_work_job_spec.rb:15 # AttachFilesToWorkJob with each of the different use files attaches the files to the work and updates them
rspec ./spec/jobs/attach_files_to_work_job_spec.rb:25 # AttachFilesToWorkJob with no use specified attaches the files to the work and updates them
```

ran the entire spec file and it was green locally, 2/2.

RudyOnRails:
Hi @adam. After speaking with the team here, can we move the checking of the `force_preferred_representation` flag to the root level of the params?

Something along the lines of `let(:params) { { force_preferred_representation: "1" } }`

awead:
@RudyOnRails Sure thing. Do you want the parameter to be "1", "true", or both? The string "1", not the number 😉

RudyOnRails:
@awead just true should be ok! If we find out `"1"`'s :wink: are needed later, I'll add them in here and in `duplicate_check`, down the road...

awead:
@RudyOnRails This should be all fixed up. Let me know if you see any other issues.

RudyOnRails:
thanks @awead, LGTM - I am going to rebase on a fancy new green develop branch, and rerun travis here, then merge in tomorrow AM.

app/controllers/lakeshore/ingest_controller.rb
```diff
@@ -6,11 +6,10 @@ class IngestController < APIController
     # load_and_authorize_resource :curation_concern, class: 'GenericWork'
 
     delegate :intermediate_file, :asset_type, :ingestor, :attributes_for_actor,
-             :check_duplicates?, :represented_resources, :force_preferred_representation?, to: :ingest
+             :check_duplicates_turned_off?, :represented_resources, :force_preferred_representation?, to: :ingest
 
     before_action :validate_ingest, :validate_asset_type, only: [:create]
     before_action :validate_duplicate_upload, :validate_preferred_representations, only: [:create, :update]
-    skip_before_action :validate_duplicate_upload, if: :duplicate_check_param_is_false
```
RudyOnRails, line 13: @awead removing this since I added the guard to our `validate_duplicate_upload` before_filter, following your guard clause logic from previous discussions...

app/controllers/lakeshore/ingest_controller.rb
```diff
@@ -42,6 +41,7 @@ def validate_asset_type
       end
 
       def validate_duplicate_upload
+        return if check_duplicates_turned_off?
```
RudyOnRails, line 44: @awead this is all I ever really needed (plus a bunch of tests I wrote) as far as logic goes for respecting the duplicate_check param. I changed the name just so it's a little more specific in what it's returning.

app/controllers/lakeshore/ingest_controller.rb
```diff
@@ -55,10 +55,6 @@ def validate_preferred_representations
 
     private
 
-      def duplicate_check_param_is_false
-        params.fetch(:duplicate_check, nil) == "false"
-      end
-
```
RudyOnRails, line 61: Didn't need this anymore as it already existed in Ingest class.

app/models/lakeshore/ingest.rb
```diff
@@ -44,9 +44,9 @@ def valid_update?
     end
 
     # @return [true, false]
-    # Only returns false if the parameter is explicitly set to "false"
-    def check_duplicates?
-      params.fetch(:duplicate_check, nil) == "false" ? false : true
+    # Only returns true if the parameter is explicitly set to "false"
+    def check_duplicates_turned_off?
+      params.fetch(:duplicate_check, nil) == "false"
```
RudyOnRails: this comparator will always return true or false, so I removed ternary. I also updated the docu.

RudyOnRails:
[push build](https://travis-ci.org/aic-collections/aicdams-lakeshore/jobs/366229542) failed due to SQLite lock error. only 1 failure, rerunning. will merge after lunch then rebase under orphaned permissions PR

PR #463, opened 2018-04-12
Orphaned ACLs from batch edit

awead:
InheritPermissionsJob was running when assets' permissions were changed
which was resulting in orphaned assets being created in Fedora outside
of the base path.

Because file sets use the same ACL resources as their parent assets,
there is no longer any need for file sets to inherit their
permissions--they already are by virtue of using the same ACLs.

Removing the call to InheritPermissionsJob fixes the orphaned ACL issue.

### QUESTIONS/TODO's
- [ ] verify this fixes the issue in staging
- [ ] delete the remaining orphaned ACLs in production once this is deployed

### DIFF
--- a/app/controllers/batch_edits_controller.rb
+++ b/app/controllers/batch_edits_controller.rb
@@ -68,16 +68,14 @@ class BatchEditsController < ApplicationController
     end
 
     # Remove (and its commit) when https://github.com/projecthydra/sufia/issues/2450 is closed
-    # Updates terms, permissions, and visibility for a given object in a batch.
-    # Note: Permissions and visibility are *always* copied down to any contained FileSet objects.
-    #       There is no UI option presented to the user to prevent this, unlike the option that
-    #       is present when changing permissions on a single work.
+    # Updates terms and visibility for a given object in a batch.
+    # Note: Visibility is copied down to any contained FileSet objects, but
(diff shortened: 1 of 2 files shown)


RudyOnRails:
failing travis which is probably b/c of random rspec seeding issue, but I'm going to rebase onto develop, as @awead confirmed this is off master b/c he was testing.

RudyOnRails:
@awead what about this call to `InheritPermissionsJob`? does LAKEshore use this permissions controller much?
`/Users/kmusio/.rvm/gems/ruby-2.2.2/gems/sufia-7.2.0/app/controllers/curation_concerns/permissions_controller.rb`
```
class CurationConcerns::PermissionsController < ApplicationController
  include CurationConcerns::PermissionsControllerBehavior

  def confirm_access
    # intentional noop to display default rails view
  end

  def copy_access
    authorize! :edit, curation_concern
    # copy visibility
    VisibilityCopyJob.perform_later(curation_concern)

    # copy permissions
    InheritPermissionsJob.perform_later(curation_concern)
    redirect_to [main_app, curation_concern], notice: I18n.t("sufia.upload.change_access_flash_message")
  end
end
```

RudyOnRails:
rebased and travis is running...

PR #469, opened 2018-04-13
.travis.yml uses hard-coded ruby version, needs to rely on .ruby-version

### LINKED ISSUE #468: .travis.yml uses hard-coded ruby version, needs to rely on .ruby-version

RudyOnRails:
resolves #468
https://cits.artic.edu/issues/2850

all environments need to use the same ruby version.

Travis currently uses "rvm": "2.3.1", from the yaml, but if you remove it will wonderfully fallback to the .ruby-version file.
https://docs.travis-ci.com/user/languages/ruby/#Using-.ruby-version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: ruby
 sudo: false
-rvm:
-  - 2.3.1
 cache:
   bundler: true
   directories:


RudyOnRails:
not waiting for travis to pass b/c this is just a version change to the version that dev/test/stg/prod/lcl run.

PR #470, opened 2018-04-25
Add "Constituent part" relationship, #2168

awead:
Add the concept of "constituent part" as a relationship between two assets.

The usage pattern, workflow handling and Fedora storage policy should be the same as "has attachment/is attachment of".

# REDMINE URL: https://cits.artic.edu/issues/2168


awead:
@RudyOnRails this is ready. Might be nice to push the branch to TEST to try it out first.

app/forms/concerns/asset_form_behaviors.rb
```diff
@@ -18,11 +18,13 @@ def self.build_permitted_params
         { representations_for: [] },
         { documents_for: [] },
         { attachments_for: [] },
+        { has_constituent_part: [] },
```
RudyOnRails, line 21: @awead just for clarification, if a complex asset has, let's say 360, individual slices/constituents, when ingesting the constituent assets, wouldn't the predicate on the parent asset be `has_constituent_parts`? Pardon the question, I may be misunderstanding this.

RudyOnRails:
@awead pushing to TEST sounds like a good idea, but @kefo said it's been broken for a bit, and I haven't had much of a chance to look into it. Any ideas? Also, is the TEST environment setup just for the relationship-flip right now? Like -- does it have a custom Fedora/Solr/DB?

awead:
@RudyOnRails I don't believe so, but we should check with @scossu

scossu:
Can we push it to dev?

RudyOnRails:
@scossu Sure we can put it on DEV. DEV currently has duplicate_check on it, I believe, with an additional column or two on uploaded_files table. Since it's _extra_ data, I don't think it will cause any problems if we deploy a different version, versus deploying code that depends on nonexistent data.

Do you happen to know if TEST is reserved for the relationship-flip currently, and has any custom Fedora or Solr or PG?

RudyOnRails:
@awead @scossu this branch has been deployed to DEV env.

app/forms/concerns/asset_form_behaviors.rb
RudyOnRails: My fault, I see that the predicate for if a work has many reps, it is named hasRepresentation.

PR #472, opened 2018-04-27
add PG mintuid stored procedure to version control

RudyOnRails:
https://cits.artic.edu/issues/2904


RudyOnRails:
PR build failed, PUSH is green. Probably need a Lakeshore restore line due to resource already existing in Fedora, and all this diff includes is adding a new txt file!:
```
  1) BatchEditsController#update when visibility is unchanged preserves the objects' original permissions
     Failure/Error: yield
     
     Ldp::Conflict:
       Can't call create on an existing resource
```

Just going to restart PR build.

PR #487, opened 2018-06-11
feature 922: Separate "depositor" from "owner"

RudyOnRails:
https://cits.artic.edu/issues/922

### DIFF
--- a/app/models/concerns/asset_metadata.rb
+++ b/app/models/concerns/asset_metadata.rb
@@ -80,6 +80,6 @@ module AssetMetadata
                      :second_document_sub_type, :publish_channels, :attachments, :copyright_representatives,
                      :licensing_restrictions, :constituent_of
 
-    property :dept_deposited, predicate: AIC.deptDeposited, multiple: false, class_name: "Department"
+    property :owner, predicate: AIC.owner, multiple: false, class_name: "AICUSer"
   end
 end
--- a/app/models/concerns/asset_metadata.rb
+++ b/app/models/concerns/asset_metadata.rb
@@ -80,8 +80,6 @@ module AssetMetadata
                      :second_document_sub_type, :publish_channels, :attachments, :copyright_representatives,
                      :licensing_restrictions, :constituent_of
 
-    property :dept_deposited, predicate: AIC.deptDeposited, multiple: false do |index|
-      index.as :stored_searchable, :facetable, using: :pref_label
-    end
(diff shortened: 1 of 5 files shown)


lib/aic.rb
```diff
@@ -81,6 +81,9 @@ class AIC < RDF::StrictVocabulary("http://definitions.artic.edu/ontology/1.0/")
   term :deptCreated,
        label: "Created by Department",
        comment: "Department of resource creator"
+  term :deptDeposited,
+       label: "Deposited by Department",
+       comment: "Department of resource depositor"
   term :digitizationSource,
```
RudyOnRails: @awead Is this file generated? I think not b/c I couldn't find a script associated with it. It's okay to add new terms directly to this?

app/models/concerns/asset_metadata.rb
```diff
@@ -79,5 +79,9 @@ def second_document_sub_type=(value)
     accepts_uris_for :keyword, :digitization_source, :document_type, :first_document_sub_type,
                      :second_document_sub_type, :publish_channels, :attachments, :copyright_representatives,
                      :licensing_restrictions, :constituent_of
+
+    property :dept_deposited, predicate: AIC.deptDeposited, multiple: false do |index|
+      index.as :stored_searchable, :facetable, using: :pref_label
+    end
   end
```
RudyOnRails: @awead does this look ok as far as just adding a new property? I wasn't sure about line 84 at all. I assume we want it stored, and searchable, and Becca wants to be able to narrow down blacklight searches with the facet. Not sure what the `using: :pref_label` is for.

lib/aic.rb
awead: Yes, that's exactly what it's for!

app/models/concerns/asset_metadata.rb
awead: @RudyOnRails Is this property another Fedora resource, or is it a string literal? I think you want a Fedora resource here.  `using: :pref_label` was a feature I added to ActiveFedora where it will solrize the resource according to the value of `pref_label`. However, it doesn't work with single-valued attributes, so you'll need to use `IndexingBehaviors.pref_label_for` and add the term to the asset indexer. Look at the `digitization_source` property as an example.

RudyOnRails: @awead sorry for delay! Yeah, it's a fedora resource. Thanks for the example, it worked well after I removed the block from the `property` call. Duh - don't need to pass a block to spec out the index details when you just plop it directly into `generate_solr_document`.

PR #493, opened 2018-06-28
Don't allow unregistered file types in the ingest API

awead:
This changes the ingest so you cannot upload arbitrary file types. We
still need to allow null file types, but this will be addressed later.

# REDMINE URL: https://cits.artic.edu/issues/2773

### QUESTIONS/TODO's
- [ ] We still need to allow null file types


awead:
@RudyOnRails Ditto here. We can merge this if develop is the current target.

PR #501, opened 2018-07-30
2420: asset not showing correct access_master after changing intermediate file set

RudyOnRails:
# REDMINE URL: https://cits.artic.edu/issues/2420


app/controllers/concerns/curation_concerns/file_sets_controller_behavior.rb
```diff
@@ -87,7 +87,7 @@ def update
                     sufia_uploaded_file = Sufia::UploadedFile.new(file: file,
                                                                   uploaded_batch_id: UploadedBatch.create.id,
                                                                   user: current_user)
-                    sufia_uploaded_file.valid? ? actor.update_content(file) : false
+                    sufia_uploaded_file.valid? ? actor.update_content(sufia_uploaded_file.file) : false
```
RudyOnRails: Let's keep this commit out of this PR, as it is unrelated.

app/prepends/prepended_controllers/with_stripped_down_file_set_id.rb
```diff
@@ -0,0 +1,5 @@
+module PrependedControllers::WithStrippedDownFileSetId
+  def image_id
+    params[:id].partition("%").first
```
RudyOnRails: This is a great workaround, but can we allow the full id through to the `Riiif::Image` model (I think this may improve Riiif cache_keys as well, because they use `id`. I believe this was the issue with mixed tiles in the UV.) and then reduce it to the file_set id inside the `riiif.rb` initializer so that `DerivativePath.access_path(id)` can still be located ok? 

```
Riiif::Image.file_resolver.id_to_uri = lambda do |id|
  id = id.partition("/").first # SOMETHING LIKE THIS
  url = DerivativePath.access_path(id)
  Rails.logger.info "Riiif resolved #{id} to #{url}"
  url
end
```

Possibly add it to `DerivativePath::access_path(object)` instead, but I'm not sure if that is always a string:

```
  class << self
    def access_path(object)
      new(object).access_path
    end
  end
```

due to the initializer doing a typeOf:
```
    def initialize(object, destination_name = nil)
      @id = object.is_a?(String) ? object : object.id
      @destination_name = destination_name.gsub(/^original_file_/, '') if destination_name
    end
```

app/presenters/file_set_presenter.rb
```diff
@@ -14,10 +14,17 @@ def role
   # Only for the access master
   def display_image
     return unless rdf_types.include?(AICType.IntermediateFileSet)
+
+    original_file = ::FileSet.find(id).original_file
+
+    latest_file_id = original_file.has_versions? ? ActiveFedora::File.uri_to_id(original_file.versions.last.uri) : original_file.id
```
RudyOnRails, line 20: This will hit Fedora as noted in the Hyrax PR, but I don't think it's too expensive. If it is, we can design another unique endpoint based on the file_set's last_modified in Solr:

```
# grab solr doc
solr_doc = ActiveFedora::SolrService.query("id:#{id}", rows: 1).map { |hit| SolrDocument.new(hit) }.first
# build presenter
fs_presenter = FileSetPresenter.new(solr_doc, current_ability)
# latest_file_id = id + "#" + fs_presenter.last_modified
```

We may get newer manifests more often since a FileSet may just have its title modified and not an entire version updated, and the newer iiif_endpoint will fire a new info.json/`info_service` request, dimensions would come back the same, and then a new `tmp/network_file` would get created due to the new `id` getting hashed.

We'd need to compare the pros and cons.

app/presenters/file_set_presenter.rb
```diff
@@ -14,10 +14,17 @@ def role
...
+
+    encoded_id = ActionDispatch::Journey::Router::Utils.escape_segment(latest_file_id)
```
RudyOnRails, line 22: Will this need to be decoded back on the other side? What's the difference between `CGI::escape` and `ActionDispatch::Journey::Router::Utils.escape_segment`?

config/initializers/riiif.rb
```diff
@@ -17,4 +17,4 @@
 Riiif::Image.authorization_service = IIIFAuthorizationService
 
 Riiif.not_found_image = 'app/assets/images/us_404.svg'
-Riiif::Engine.config.cache_duration_in_days = 365
+Riiif::Engine.config.cache_duration_in_days = 0
```
RudyOnRails: Not 100% what the result of this will be - Adam and I were dialing a few different settings. Is this basically turning caching off Riiif caching since all `Rails.cache::fetch` calls will use an `expire_in` of 0 seconds?

RudyOnRails:
Also need to get two failing specs to pass by stubbing the newly added ::FileSet.find's:

```
     ActiveFedora::ObjectNotFoundError:
       Couldn't find FileSet with 'id'=too-big
```

PR #507, opened 2018-08-03
3015: API&GUI: when assets are created by Imaging Dept, requestors have no access to them

RudyOnRails:
# REDMINE URL: https://cits.artic.edu/issues/3015

### DIFF
--- a/app/models/generic_work.rb
+++ b/app/models/generic_work.rb
@@ -17,7 +17,12 @@ class GenericWork < Resource
   type type + aic_type
 
   before_create :status_is_active, :public_is_false
+  after_create :give_on_behalf_of_dept_write_permission, if: :on_behalf_of?
   validate :id_matches_uid_checksum, on: :update
+
+  def on_behalf_of?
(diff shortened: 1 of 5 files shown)


app/models/generic_work.rb
```diff
@@ -85,6 +90,14 @@ def legacy_file_set
     members.select { |f| f.type.include?(AICType.LegacyFileSet) }
   end
 
+  def give_on_behalf_of_dept_write_permission
+    u = User.find_by_email(on_behalf_of)
+    groups = edit_groups
+    groups << u.department
+    self.edit_groups = groups
+    save
```
RudyOnRails, line 98: @awead I know I had the idea of doing this in an actor as well. It was working okay this way, but I wonder if on all new asset Phoenix requests (that will include the new on_behalf_of param), this will do an unnecessary save and create Fedora messaging overhead?

If I were to do this in the actor stack, would I not have to explicitly save, but just pass the env on to the `next_actor` and somewhere else in the stack it would get saved?

PR #513, opened 2018-08-08
3055: Deleted asset cannot be re-uploaded (upload area dup warning)

RudyOnRails:
# REDMINE URL: https://cits.artic.edu/issues/3055


app/jobs/concerns/master_derivatives_job.rb
```diff
@@ -0,0 +1,9 @@
+class MasterDerivativesJob < ActiveJob::Base
```
RudyOnRails, line 2: @awead is this cool? I didn't want to dupe the code in each job, so I subclassed each job from a new superclass and added in the after_perform.

Not quite sure why we have duplicate derivatives jobs either, but maybe that's for another day's discussion.

PR #519, opened 2018-08-22
2963.1 create function of create_or_update endpoint

RudyOnRails:
# REDMINE URL: https://cits.artic.edu/issues/2963

### QUESTIONS/TODO's
- [ ] 

### SUPPORTING INFO
* This is the remaining "create" side of the create_or_update endpoint.

### DIFF
--- a/app/controllers/concerns/curation_concerns/file_sets_controller_behavior.rb
+++ b/app/controllers/concerns/curation_concerns/file_sets_controller_behavior.rb
@@ -191,7 +191,7 @@ module CurationConcerns
       end
 
       def empty_file?(file)
-        (file.respond_to?(:tempfile) && file.tempfile.empty?) || (file.respond_to?(:size) && file.empty?)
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/curation_concerns/file_sets_controller_behavior.rb
```diff
@@ -191,7 +191,7 @@ def json_error(error, name = nil, additional_arguments = {})
       end
 
       def empty_file?(file)
-        (file.respond_to?(:tempfile) && file.tempfile.empty?) || (file.respond_to?(:size) && file.empty?)
+        (file.respond_to?(:tempfile) && file.tempfile.blank?) || (file.respond_to?(:size) && file.blank?)
```
RudyOnRails, line 194: @awead does this make sense? I was getting some `NoMethodErrors` for `#empty?`

awead: @RudyOnRails yes, I often use `present?` and `blank?` because they handle arrays that are either `[]` or `nil`.

PR #521, opened 2018-08-27
3071: user can still get scrambled UV tiles if they refresh asset show page before new version's access-master deriv is complete

RudyOnRails:
# REDMINE URL: https://cits.artic.edu/issues/3071

### QUESTIONS/TODO's
- [ ]

### SUPPORTING INFO
*

### DIFF
--- a/app/presenters/file_set_presenter.rb
+++ b/app/presenters/file_set_presenter.rb
@@ -13,11 +13,14 @@ class FileSetPresenter < Sufia::FileSetPresenter
   # @return [IIIFManifest::DisplayImage]
   # Only for the access master
   def display_image
+    # intermediate file_sets only
     return unless rdf_types.include?(AICType.IntermediateFileSet)
 
-    original_file = ::FileSet.find(id).original_file
+    access_master_path = DerivativePath.access_path(id)
 
-    latest_file_id = original_file.has_versions? ? ActiveFedora::File.uri_to_id(original_file.versions.last.uri) : original_file.id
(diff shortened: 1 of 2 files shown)


app/presenters/file_set_presenter.rb
```diff
@@ -13,11 +13,14 @@ def role
...
+    # intermediate file_sets only
     return unless rdf_types.include?(AICType.IntermediateFileSet)
 
-    original_file = ::FileSet.find(id).original_file
+    access_master_path = DerivativePath.access_path(id)
 
-    latest_file_id = original_file.has_versions? ? ActiveFedora::File.uri_to_id(original_file.versions.last.uri) : original_file.id
+    modified_time_of_access_master = File.mtime(access_master_path).to_s
+
+    latest_file_id = "#{id}/#{modified_time_of_access_master}"
```
RudyOnRails, line 23: @awead an idea came to me while 40,000ft up...just use the modified time of the access_master deriv, same way we customized the `HTTPFileResolver`. That's easier than trying to udpate the `system_modified_` from an after_perform of the derivatives job. Plus this way modifications (to title etc.) to a file_set won't trigger a new manifest...only when the derivative jp2 has a new modified time.

awead: @RudyOnRails Yep, that should work. I wonder if hitting disk each time the page loads is a problem. Could you index the access master's modification time in the file set? Definitely for later though. We just got done reindexing file sets...

PR #529, opened 2018-09-14
2352 add rows param to Solr query

RudyOnRails:
# REDMINE URL: https://cits.artic.edu/issues/2352

### QUESTIONS/TODO's
- [ ] we should do this high level stuff, or figure out these high level things

### SUPPORTING INFO
*

### DIFF
--- a/app/models/aic_user.rb
+++ b/app/models/aic_user.rb
@@ -63,7 +63,7 @@ class AICUser < CitiResource
 
       def find_nick_by_solr(nick)
         return nil unless nick
-        docs = ActiveFedora::SolrService.query("nick_ssim:#{nick}", fq: 'has_model_ssim:"AICUser"')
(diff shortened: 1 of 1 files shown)


app/models/aic_user.rb
```diff
@@ -63,7 +63,7 @@ def json_result(doc)
 
       def find_nick_by_solr(nick)
         return nil unless nick
-        docs = ActiveFedora::SolrService.query("nick_ssim:#{nick}", fq: 'has_model_ssim:"AICUser"')
+        docs = ActiveFedora::SolrService.query("nick_ssim:#{nick}", rows: 1, fq: 'has_model_ssim:"AICUser"')
```
awead, line 66: 🤦‍♂️

RudyOnRails: :man_shrugging:

PR #533, opened 2018-10-06
3151/2520: Provide facilities/workflow for users to delete an asset that is a pref rep

RudyOnRails:
# REDMINE URL: https://cits.artic.edu/issues/3151

### QUESTIONS/TODO's
- [ ] we should do this high level stuff, or figure out these high level things

### SUPPORTING INFO
* still need to do batch, and removing a preferred CR from the relationships tab, but those will be in next release.


RudyOnRails:
closing since I already brought in here: https://github.com/aic-collections/aicdams-lakeshore/pull/534

